PR #1, opened 2020-12-01
MCS-449 through MCS-460:  Getting UI to a solid state

brianpippin:
All work in tickets are completed, but UI still a work in progress.  Rest of tickets should finish it off.


mcs_docker_setup/analysis-ui/src/components/App/home.jsx
```diff
@@ -0,0 +1,101 @@
...
+                        if (loading) return <div>No stats yet</div> 
+                        if (error) return <div>Error</div>
+                        
+                        console.log(data[homeStatsQuery]);
+                        let homeStats = data[homeStatsQuery];
+
+                        // Values For By Scores Chart
+                        const byScoreLength = homeStats.statsByScore.length - 1;
+                        let byScoreKeys = Object.keys(homeStats.statsByScore[byScoreLength]);
+                        byScoreKeys.shift();
```
rartiss55, line 50: Maybe its just too late in the day for me but why do we need to do shift() for these and why is the byScoreMaxValue math a little different from interactiveMaxValue and passiveMaxValue?

mcs_docker_setup/docker-compose.yml
```diff
@@ -50,77 +50,49 @@ services:
...
-    healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:9200"]
-      interval: 1s
-      timeout: 1s
-      retries: 30
-  neon-server:
     environment:
-      STATE_DIRECTORY: /config
-      SERVER_PORT: 80
+      - MONGO_INITDB_DATABASE=mcs
```
rartiss55, line 92: I think you mentioned this already at least for the AWS credentials but wanted to flag for later the database usernames/passwords here + in mcs_ingest as something we should potentially extract out of git later and store elsewhere.

brianpippin: I'm slightly less worried about these but you still make an excellent point, the only thing i'm slightly less worried about is someone would need to have admin access to our AWS to be able to even get to the mongo database.  Where as the aws credentials would allow all kinds of mayhem.

mcs_docker_setup/analysis-ui/src/components/App/home.jsx
brianpippin: So the chart component I am using is:  https://nivo.rocks/bar - you need to pass in the keys along with the chartIndex.  keys={keys} 
indexBy={chartIndex}
When the data comes back from the server one item looks like {"test_type": "shape_constancy", "ibm": 10, "opics": 10, "mess": 10} The test_type is the index - it is always first and might be a different field value.  The rest is the preformers, so for the keys (which is how the chart does the groupBy and the legend) we want all the field values except the first one, which is also then the chart index.  Perhaps a comment in the code explaining this would be useful.

mcs_docker_setup/node-graphql/server.statsFunctions.js
```diff
@@ -0,0 +1,123 @@
...
+}
+
+const statsByScore = function(scoreStats){
+    scoreStats.sort((a, b) => (a._id.performer > b._id.performer) ? 1 : -1);
+
+    let scoreOverallCorrect = {"score_type": "Overall Correct"};
+    let scoreOverallIncorrect = {"score_type": "Overall Incorrect"};
+    let scorePlausibleCorrect = {"score_type": "Plausible Correct"};
+    let scorePlausibleIncorrect = {"score_type": "Plausible Incorrect"};
+    let scoreImpluasibleCorrect = {"score_type": "Implausible Correct"};
```
rartiss55, line 23: typo here with scoreImpluasibleCorrect

mcs_docker_setup/resources/htpasswd
```diff
@@ -0,0 +1 @@
+mcsEval:$2y$05$W1H.84LmC4WJlVchpQ957eo17QrxFWXK.0BlJKfO6NCObq1edUOuC
```
rartiss55, line 1: Hmm should we store this here?

mcs_docker_setup/docker-compose.yml
rartiss55: Makes sense -- I added a ticket to the backlog so that we can come back to this later (MCS-469).

rartiss55:
Left a couple more comments about things but just wanted to say I think this looks great so far and I'm excited to see the new UI!

mcs_docker_setup/resources/htpasswd
brianpippin: I don't even know that I am using this file now that I see this, I will need to check but this might be OBE.

mcs_docker_setup/analysis-ui/src/components/App/home.jsx
brianpippin: Also I forgot to add, the way that it comes back from the server is also exactly how 'data={data}' portion of the component expects the data, so it's just manipulating the array of data that is precisely how we need it, to get the 'keys' field.  Just some additional context.

mcs_docker_setup/resources/htpasswd
brianpippin: Oddly enough this file is in the neon-jan-2020.tar.gz and just never got commited back into the resources folder.  The docker compose files using it for nginx - ./resources/htpasswd:/etc/nginx/htpasswd:ro

PR #4, opened 2020-12-14
MCS-465: Design for tabs and saving and loading saved queries

brianpippin:


mcs_docker_setup/docker-compose-dev.yml
```diff
@@ -15,7 +15,6 @@ services:
       - ./dist:/var/www:ro
       - ./resources/nginx.conf:/etc/nginx/conf.d/default.conf:ro
       - ./resources/htpasswd:/etc/nginx/htpasswd:ro
-      - ./resources/config.yaml:/var/www/app/config/config.yaml
```
rartiss55, line 18: We should probably remove this line from docker-compose-prod.yml too, right?

brianpippin: docker-compose.yml is the prod one.  It's already in the code review.

mcs_docker_setup/analysis-ui/src/css/app.css
```diff
@@ -607,4 +607,91 @@ ol, ul {
 
 .toggle-percent-number-container {
     margin-left: auto;
+}
+
+.load-query-contents {
```
rartiss55, line 612: More of a thing to consider for later -- would we want to start making separate CSS files as the app grows?

rartiss55 (approved):
Looks good, and thanks for leaving the TODO item for MCS-463 -- I think that was my only question since it didn't look like reloading saved queries into the UI/having the drop downs update was working.

mcs_docker_setup/analysis-ui/src/css/app.css
brianpippin: Yes I saw your code review and thought, we should be moving to separate CSS files, we should transition to that as we keep working on things.

brianpippin:
> Looks good, and thanks for leaving the TODO item for MCS-463 -- I think that was my only question since it didn't look like reloading saved queries into the UI/having the drop downs update was working.

Yeah I already have all of that working, but it was a big refactor for the query builder to move states up a level and redesign that component it didn't make sense to do the work twice.

PR #5, opened 2020-12-15
MCS-458: Add ability to change password to 'My Account' page

rartiss55:
Also added placeholder disabled inputs for now for username/email.

@brianpippin let me know if I missed anything for this ticket.


brianpippin:
I think this looks great, the only thing I noticed is, I think you should have something reset the password fields to empty after the password is updated.  Currently the text blocks just keep the current and new passwords in there.  I think if the password update is successful you should empty string them out.

rartiss55:
Good idea @brianpippin -- I updated those to clear on success.

brianpippin (approved):

PR #6, opened 2020-12-21
MCS-463: Complex Query Builder

brianpippin:
Updated the UI for adding parameters to a query and removing, and storing query objects.


brianpippin:
So I accidentally pushed the table ticket to this code review so this now a 2 for 1.  Sorry about the large code block.

rartiss55 (approved):
Looks good -- did notice the queries could be a bit slow, but that might just be my machine vs an actual production environment.

PR #7, opened 2020-12-30
MCS-490, MCS-491, MCS-498 - Fix Analysis page URL params and dropdowns, add link to analysis UI from query results table

rartiss55:
Now the page won't reload when new dropdown selections are made, making selections will run the correct queries, and the URL params will stay consistent with the underlying state.


brianpippin (approved):
I still need to checkout and test it but the code looks good.

PR #10, opened 2021-01-14
MCS-508: Field field drop downs in query builder based on collection

brianpippin:
Yes I know i mistyped the branch as MC-508 i'll just delete it after merging :)


rartiss55:
Code looks good -- the only thing that might come up later is that our eval names have to include "Results" or "Scenes" at the moment (I still have some old eval 2 data that this broke on), but this won't be an issue in Prod.

rartiss55 (approved):

PR #12, opened 2021-01-14
MCS-515 - Update table to use same table component that is on query builder page

rartiss55:
Had some issues with this initially, and ended up using the QueryResultsTable component itself and trying to generalize things a bit more. May have not been worth it (if not, I can create a new table component for the scenes view instead).

I also messed with the fields/labels we're displaying in the scores table, and added back in the complexQuery to display the slices data within the scores table.


rartiss55:
Closing -- new PR here (without changes to QueryResultsTable) https://github.com/NextCenturyCorporation/mcs-ui/pull/15

PR #13, opened 2021-01-15
MCS-525: Make the select a multiselect for the query builder

brianpippin:


deanwetherby (approved):

mcs_docker_setup/node-graphql/server.mongoSyntax.js
```diff
@@ -40,10 +40,12 @@ const createComplexMongoQuery = function(queryObj){
 
         switch(queryObj[i]["functionOperator"]) {
             case EQUALS:
+                const inObjList = queryObj[i]["fieldValue1"].split(",");
```
ThomasSchellenbergNextCentury, line 43: Are there any strings that aren't part of lists that have commas? I can't think of any offhand, but if there are, that would screw this up, right? I'm not sure if it's easy for you to run a query to double-check it.

ThomasSchellenbergNextCentury: Comma is *probably* fine. Anything else we choose (like semicolon or whatever) could have the same exact problem. You could also plan on stripping all commas (or whatever delimeter that you want to use) from all strings the next time you reingest the data.

Can't mongo have array-type columns?

If not, in the future, one option is to convert the contents of ALL the cells into JSON strings.

For now, given the timeline, I'm fine if you just want to deal with this later.

brianpippin: @ThomasSchellenbergNextCentury - I changed this list to be separated by "__,__" and wrote the code to use that instead.  We should have to worry about that combination in the data :)

ThomasSchellenbergNextCentury (approved):

PR #15, opened 2021-01-19
MCS-515 - Add sorting to scores table

rartiss55:
Opening new PR for MCS-515. Scores table is using Material UI and now has sorting. 

I also messed with the fields/labels we're displaying in the scores table (removed MSE and renamed 'Answer' to 'Classification'), and added back in the complexQuery to display the slices data within the scores table.

### DIFF
--- a/mcs_docker_setup/analysis-ui/src/components/App/scenesEval3.jsx
+++ b/mcs_docker_setup/analysis-ui/src/components/App/scenesEval3.jsx
@@ -5,6 +5,12 @@ import _ from "lodash";
 import $ from 'jquery';
 //import FlagCheckboxMutation from './flagCheckboxMutation';
 import {EvalConstants} from './evalConstants';
+import Table from "@material-ui/core/Table";
+import TableBody from "@material-ui/core/TableBody";
+import TableCell from "@material-ui/core/TableCell";
+import TableHead from "@material-ui/core/TableHead";
+import TableRow from "@material-ui/core/TableRow";
+import TableSortLabel from "@material-ui/core/TableSortLabel";
 
 const historyQueryName = "getEval3History";
 const sceneQueryName = "getEval3Scene";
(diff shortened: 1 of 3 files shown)


mcs_docker_setup/analysis-ui/src/components/App/scenesEval3.jsx
```diff
@@ -310,18 +370,31 @@ class ScenesEval3 extends React.Component {
...
-                {   "categoryType":  this.props.value.category_type,
-                    "sceneNum": parseInt(this.props.value.scene_num)
+            <Query query={create_complex_query} variables={
+                {    
+                    "queryObject":  this.getSceneHistoryQueryObject(
+                        this.props.value.category_type,
+                        this.props.value.scene_num
+                    ), 
+                    "projectionObject": projectionObject
                 }} fetchPolicy='network-only'>
```
brianpippin, line 391: I would consider removing this fetchPolicy, if you said it to network it will make the call every time the page needs to rerender, however, I believe we can cache this call because none of the data will be changing.

mcs_docker_setup/analysis-ui/src/components/App/scenesEval3.jsx
```diff
@@ -310,18 +370,31 @@ class ScenesEval3 extends React.Component {
...
+            sortBy: property 
+        });
+    };
+
     render() {
         return (
-            <Query query={mcs_history} variables={
-                {   "categoryType":  this.props.value.category_type,
-                    "sceneNum": parseInt(this.props.value.scene_num)
+            <Query query={create_complex_query} variables={
```
brianpippin, line 384: I also think you could have saved a lot of work by creating a custom endpoint for this call, and just manually build the query on the backend, but I think this is a useful paradigm as well.

mcs_docker_setup/node-graphql/server.mongoSyntax.js
```diff
@@ -40,16 +40,29 @@ const createComplexMongoQuery = function(queryObj){
 
         switch(queryObj[i]["functionOperator"]) {
             case EQUALS:
-                const inObjList = queryObj[i]["fieldValue1"].split("__,__");
+                let inObjList;
+                if(isNaN(queryObj[i]["fieldValue1"])) {
```
brianpippin, line 44: Only attempt to split for multivalue if it's a string.

mcs_docker_setup/node-graphql/server.mongoSyntax.js
```diff
@@ -40,16 +40,29 @@ const createComplexMongoQuery = function(queryObj){
...
-                        inObjList[i] = false;
+                    if(isNaN(inObjList[i])){
+                        if(inObjList[i].toLowerCase() === "true") {
+                            inObjList[i] = true;
+                        }
+                        if(inObjList[i].toLowerCase() === "false") {
+                            inObjList[i] = false;
+                        }
+                    } else {
+                        if(parseFloat(inObjList[i]) !== NaN) {
```
brianpippin, line 59: Since we are already doing $in, we can check for a 2 or "2" in our query and we will find either value depending on which is in the field.

brianpippin (approved):

PR #16, opened 2021-01-19
MCS-526:  Agency Pair Scoring

brianpippin:


brianpippin:
Also a fix to remove % from the grouped by, along with making only the arrow clickable in the grouping row.  Changes requested by Psychologists.

rartiss55 (approved):

PR #19, opened 2021-01-26
MCS-538: Fix javascript crash when using 'x' to remove item from multiselect

brianpippin:


mcs_docker_setup/analysis-ui/src/components/App/queryLine.jsx
```diff
@@ -111,10 +111,12 @@ class QueryLineItem extends React.Component {
 
     selectDropDownValue = (event) => {
         let valueToSet = "";
-        for(let i=0; i < event.length; i ++){
-            valueToSet = valueToSet + event[i].value;
-            if(i < event.length -1) {
-                valueToSet = valueToSet + "__,__"
+        if(event !== null & event !== undefined){
```
brianpippin, line 114: Just added an if check for the event being null or undefined, it will set the fieldValue to ""

rartiss55: Did you mean to use the bitwise AND here?

brianpippin: A typo that still worked when I tested it.

rartiss55 (approved):

PR #20, opened 2021-01-28
MCS-514 - Fix analysis UI URLs/test vs scene references

rartiss55:
- Eval name is no longer hardcoded from queryResultsTable
- Corrected scene_num parameter to "test_num" in URL and added "scene" parameter to allow for selecting a particular scene via links

This is still a bit goofy because:
1.) The scene_num fields still need to be fixed in ingest (I created another ticket for that somewhere and I thought it was safer not to try to do this all at once)
2.) The way people want to select and view different tests + scenes/manage that state will probably factor into some analysis UI page redesign at some point since it is currently spread around about 3 or 4 components right now, which isn’t ideal IMO.

### DIFF
--- a/mcs_docker_setup/analysis-ui/src/components/App/scenes.jsx
+++ b/mcs_docker_setup/analysis-ui/src/components/App/scenes.jsx
@@ -66,7 +66,7 @@ class Scenes extends React.Component {
             flagRemove: false,
             flagInterest: false,
             testType: props.value.test_type,
-            sceneNum: props.value.scene_num
+            sceneNum: props.value.test_num
         };
     }
 
@@ -224,7 +224,7 @@ class Scenes extends React.Component {
         return (
             <Query query={mcs_history} variables={
                 {"testType": this.props.value.test_type, 
-                "sceneNum": this.props.value.scene_num
+                "sceneNum": this.props.value.test_num
                 }}>
             {
                 ({ loading, error, data }) => {
@@ -246,7 +246,7 @@ class Scenes extends React.Component {
                         return (
                             <Query query={mcs_scene} variables={
                                 {"testType": this.props.value.test_type, 
-                                "sceneNum": this.props.value.scene_num
+                                "sceneNum": this.props.value.test_num
                                 }}>
                             {
                                 ({ loading, error, data }) => {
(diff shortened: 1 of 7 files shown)


mcs_docker_setup/analysis-ui/src/components/App/scenes.jsx
```diff
@@ -61,12 +61,12 @@ class Scenes extends React.Component {
         this.state = {
```
rartiss55, line 61: Wanted to apply same logic to eval 2 code.

mcs_docker_setup/analysis-ui/src/components/App/scenes.jsx
```diff
@@ -325,56 +348,63 @@ class Scenes extends React.Component {
...
-                                                        <ul className="nav nav-tabs">
-                                                            {scenesInOrder[this.state.currentSceneNum].objects.map((sceneObject, key) => 
-                                                                <li className="nav-item" key={'object_tab_' + key}>
-                                                                    <button id={'object_button_' + key} className={key === 0 ? 'nav-link active' : 'nav-link'} onClick={() => this.changeObjectDisplay(key)}>{this.findObjectTabName(sceneObject)}</button>
-                                                                </li>
-                                                            )}
-                                                        </ul>
-                                                    </div>
-                                                    <div className="object-contents">
+                                                {scenesInOrder && this.state.currentSceneNum < scenesInOrder.length && scenesInOrder[this.state.currentSceneNum] &&
```
rartiss55, line 370: Not too much was changed here -- I think it was addressed by previous changes, but tried to fix up the property checks here because of recent discussion regarding errors in Teams just in case.

mcs_docker_setup/analysis-ui/src/components/App/scenesEval3.jsx
```diff
@@ -649,42 +670,44 @@ class ScenesEval3 extends React.Component {
...
-                                                                <table>
-                                                                    <tbody>
-                                                                        {scenesInOrder && scenesInOrder[this.state.currentSceneNum] && Object.keys(scenesInOrder[this.state.currentSceneNum].objects[this.state.currentObjectNum]).map((objectKey, key) => 
-                                                                            <tr key={'object_tab_' + key}>
-                                                                                <td className="bold-label">{objectKey}:</td>
-                                                                                <td className="scene-text">{this.convertValueToString(scenesInOrder[this.state.currentSceneNum].objects[this.state.currentObjectNum][objectKey])}</td>
-                                                                            </tr>
+                                                {scenesInOrder && this.state.currentSceneNum < scenesInOrder.length && scenesInOrder[this.state.currentSceneNum] &&
```
rartiss55, line 673: (see previous comment) adding additional property checks.

mcs_docker_setup/node-graphql/server.fieldMappings.js
```diff
@@ -27,7 +27,6 @@ const historyExcludeFields = [
 
 const historyExcludeFieldsTable = [
     "_id",
-    "eval",
```
rartiss55, line 30: Need this to build the URL from queryResultsTable to analysis page.

brianpippin: This may break the query builder, because it's going to show up as a field they can query when they already selected it, but we would need to check that doing that doesn't break anything.

rartiss55: I think that this just includes the "eval" column in the results projection but doesn't actually let you query on it, and it didn't look like the "eval" column was available as a selections under "Evaluation Results".

brianpippin (approved):
I think this is good, we should probably just get it merged, so we can then incorporate some of this, when refactoring the analysis page.

PR #21, opened 2021-02-09
MCS-543: File restructure

brianpippin:


rartiss55 (approved):

rartiss55:
Ran into an issue with not having account-configs.js in my setup properly -- should we keep a dummy version of that around? (looks like it got deleted)

brianpippin:
I added it to gitignore, you should be able to pull the one from prod if
you need it.

rartiss55 (approved):

PR #22, opened 2021-02-23
MCS-520 - UI updates to existing fields in Mongo

rartiss55:
Addresses some of the TODOs left during Eval 3 UI development. Goes with ingest changes made here to consolidate various scene identifier fields to test_num and scene_num: https://github.com/NextCenturyCorporation/mcs-ingest/pull/5


analysis-ui/src/components/Analysis/scenes.jsx
```diff
@@ -255,29 +262,29 @@ class Scenes extends React.Component {
                     if (error) return <div>Error</div>
                     
                     const evals = data[historyQueryName];
-                    console.log(evals);
-                    let scenesByPerformer = _.sortBy(evals, "scene_part_num");
+                    //console.log(evals);
```
brianpippin, line 265: Should we remove commented out console.logs ?

analysis-ui/src/components/Analysis/scenes.jsx
```diff
@@ -242,11 +242,18 @@ class Scenes extends React.Component {
         }
     }
 
+    // Switching types for testNum, so need to pad them to match existing 
+    // file names
+    // TODO: should we rename eval 2 files?
```
brianpippin, line 247: I don't know that we should rename the files, because that would be a lot of work, so I think this is fine for now, especially since we have so much to do, but I would leave the comments in.

brianpippin (approved):
Nothing in the code sticks out to me, unfortunately a lot of the comments and scene code is probably going to be totally rewritten.  I think we should rebuild it from the ground up using the new mockups.  So ideally we start fresh, copy the code we can reuse, and build it a little better.

analysis-ui/src/components/Analysis/scenes.jsx
rartiss55: That's fine - I thought so too but thought I'd put it out there. Removed the TODO.

analysis-ui/src/components/Analysis/scenes.jsx
rartiss55: done

PR #24, opened 2021-03-11
MCS-590

rartiss55:
Fixing bug with selecting scenes from analysis page table when the table is sorted by something other than the default (videos for interactive/passive step by step info wasn't always lining up). 

Also tacked on a few things for Eval 3.5.

### DIFF
--- a/analysis-ui/src/components/Analysis/evalConstants.jsx
+++ b/analysis-ui/src/components/Analysis/evalConstants.jsx
@@ -6,6 +6,7 @@ const Eval2 = {
     sceneBucket: "https://evaluation-images.s3.amazonaws.com/eval-2-scenes/",
     sceneExtension: "-debug.json",
 
+    // note that this performerPrefixMapping is only used on the Eval 2 Results Page
     performerPrefixMapping: {
         "IBM-MIT-Harvard-Stanford": "mitibm_",
         "OPICS (OSU, UU, NYU)": "opics_",
@@ -21,15 +22,20 @@ const Eval3 = {
     movieExtension: ".mp4",
     sceneBucket: "https://evaluation-images.s3.amazonaws.com/eval-3-scenes/",
     sceneExtension: "_debug.json",
-
-    performerPrefixMapping: {
-        "IBM-MIT-Harvard-Stanford": "mit_",
-        "OPICS (OSU, UU, NYU)": "opics_",
-        "MESS-UCBerkeley": "mess_",
-    }
 }
 
+const Eval3_5 = {
+    moviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.5",
+    interactiveMoviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.5/",
+    topDownMoviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.5/",
+    movieExtension: ".mp4",
+    sceneBucket: "https://evaluation-images.s3.amazonaws.com/eval-scenes-3.5/", 
+    sceneExtension: "_debug.json",
+}
+
+// TODO: MCS-589: move to ingest/store in mongo
 export const EvalConstants = {
-    "Eval2": Eval2, 
-    "Eval3": Eval3
+    "Evaluation 2 Results": Eval2, 
+    "Evaluation 3 Results": Eval3,
+    "Evaluation 3.5 Results": Eval3_5
 }
--- a/analysis-ui/src/components/Analysis/scenesEval3.jsx
+++ b/analysis-ui/src/components/Analysis/scenesEval3.jsx
@@ -94,7 +94,7 @@ class ScenesEval3 extends React.Component {
         this.state = {
             currentPerformer: props.value.performer !== undefined ? props.value.performer : "",
             currentMetadataLevel: props.value.metadata_lvl !== undefined ? props.value.metadata_lvl : "",
-            currentSceneNum: (props.value.scene !== undefined && props.value.scene !== null) ? parseInt(props.value.scene) - 1 : 0,
+            currentSceneNum: (props.value.scene !== undefined && props.value.scene !== null) ? parseInt(props.value.scene) : 1,
             currentObjectNum: 0,
             //flagRemove: false,
             //flagInterest: false,
@@ -142,7 +142,7 @@ class ScenesEval3 extends React.Component {
             let searchString = this.props.value.history.location.search;
     
             let sceneStringIndex = searchString.indexOf("&scene=");
-            let sceneToUpdate = "&scene=" + parseInt(sceneNum + 1);
+            let sceneToUpdate = "&scene=" + sceneNum;
     
             if(sceneStringIndex > -1) {
                 let newSearchString = searchString.substring(0, sceneStringIndex);
@@ -156,7 +156,7 @@ class ScenesEval3 extends React.Component {
                     search: searchString + sceneToUpdate
                 });
             }
-            this.props.updateHandler("scene", parseInt(sceneNum + 1));
+            this.props.updateHandler("scene", sceneNum);
         }
     }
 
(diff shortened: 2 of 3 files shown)


analysis-ui/src/components/Analysis/evalConstants.jsx
```diff
@@ -21,15 +22,20 @@ const Eval3 = {
...
+const Eval3_5 = {
+    moviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.5",
+    interactiveMoviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.5/",
+    topDownMoviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.5/",
+    movieExtension: ".mp4",
+    sceneBucket: "https://evaluation-images.s3.amazonaws.com/eval-scenes-3.5/", 
+    sceneExtension: "_debug.json",
 }
 
+// TODO: MCS-589: move to ingest/store in mongo
```
rartiss55, line 36: Figured it wasn't hard to add this for now until we get to this ticket, plus I had to fix something additional Eval 3-specific in the scenesEval3 file anyway (which, yes, should be renamed at some point :))

analysis-ui/src/components/Analysis/evalConstants.jsx
```diff
@@ -21,15 +22,20 @@ const Eval3 = {
...
 export const EvalConstants = {
-    "Eval2": Eval2, 
-    "Eval3": Eval3
+    "Evaluation 2 Results": Eval2, 
+    "Evaluation 3 Results": Eval3,
+    "Evaluation 3.5 Results": Eval3_5
```
rartiss55, line 40: Assuming we're sticking to this naming convention.

analysis-ui/src/components/Analysis/scenesEval3.jsx
```diff
@@ -437,16 +444,22 @@ class ScenesEval3 extends React.Component {
...
                     this.setInitialMetadataLevel(initialMetadataLevel);
 
-                    let scenesByPerformer = _.reduce(scenesByMetadata, function(result, value, key) {
-                        result[key] = _.groupBy(value, "performer");                     
-                        return result;
+                    let scenesByPerformer = _.reduce(scenesByMetadata, function(mapByMetadata, metadataValue, metadataKey) {
+                        mapByMetadata[metadataKey] = _.reduce(_.groupBy(metadataValue, "performer"), function(mapByPerformer, perfValue, perfKey) {
+                            mapByPerformer[perfKey] = _.reduce(perfValue, function(mapByScene, sceneValue) {
+                                // scene history records are now a map with scene_num as the key instead of an array
+                                mapByScene[sceneValue.scene_num.toString()] = sceneValue;
```
rartiss55, line 451: Changed the scenes within here to a map instead of array, because there are occasionally scenes missing from the history (likely due to some error during eval) and that could also throw the count off if using an array index.

analysis-ui/src/components/Analysis/scenesEval3.jsx
```diff
@@ -550,14 +563,14 @@ class ScenesEval3 extends React.Component {
                                                             </TableRow>
                                                         </TableHead>
                                                         <TableBody>
-                                                        {this.checkIfScenesExist(scenesByPerformer) && scenesByPerformer[this.state.currentMetadataLevel][this.state.currentPerformer].sort(getSorting(this.state.sortOrder, this.state.sortBy)).map((scoreObj, rowKey) => 
+                                                        {this.checkIfScenesExist(scenesByPerformer) && _.values(scenesByPerformer[this.state.currentMetadataLevel][this.state.currentPerformer]).sort(getSorting(this.state.sortOrder, this.state.sortBy)).map((scoreObj, rowKey) =>
```
rartiss55, line 566: Now a map anyway, but scenes are converted to a separate array for table sorting.

analysis-ui/src/components/Analysis/scenesEval3.jsx
```diff
@@ -306,11 +306,11 @@ class ScenesEval3 extends React.Component {
...
     }
 
-    getSceneHistoryQueryObject = (categoryType, testNum) => {
+    getSceneHistoryQueryObject = (evalName, categoryType, testNum) => {
         return [
             {
-                fieldType: "mcs_history.Evaluation 3 Results",
-                fieldTypeLabel: "Evaluation 3 Results",
+                fieldType: "mcs_history." + evalName,
+                fieldTypeLabel: evalName,
```
rartiss55, line 313: Found another leftover hardcoded eval 3 thing -- this now takes the eval property passed in from the navigation dropdown.

analysis-ui/src/components/Analysis/evalConstants.jsx
```diff
@@ -6,6 +6,7 @@ const Eval2 = {
     sceneBucket: "https://evaluation-images.s3.amazonaws.com/eval-2-scenes/",
     sceneExtension: "-debug.json",
 
+    // note that this performerPrefixMapping is only used on the Eval 2 Results Page
```
brianpippin, line 9: What would it take to remove this mapping from the UI code, would we have to go back and rename the files?  I genuinely don't remember and am just curious if it would be worth it, although this code isn't too bad for legacy items.  Just thinking about when we port some of this code to the new design.

analysis-ui/src/components/Analysis/evalConstants.jsx
brianpippin: This is great, I think we do want to figure out how to add these configs to the database and potentially have an admin page that can add/modify the constants, so we don't need to push out a new UI for it.  Maybe a future ticket if we don't already have it, I think we discussed it before.

analysis-ui/src/components/Analysis/scenesEval3.jsx
```diff
@@ -369,7 +369,8 @@ class ScenesEval3 extends React.Component {
 
     getSceneHistoryItem = (scenesByPerformer) => {
         if(scenesByPerformer !== undefined && scenesByPerformer !== null) {
-            return scenesByPerformer[this.state.currentMetadataLevel][this.state.currentPerformer][this.state.currentSceneNum];
+            let sceneNumAsString = this.state.currentSceneNum.toString();
```
brianpippin, line 372: super nitpicky and something we should look at more, but if you aren't going to modify the variable we should probably set it as const and let is used for variables we modify.

brianpippin (approved):
Besides the few minor comments I made, I think it looks super good.

analysis-ui/src/components/Analysis/evalConstants.jsx
```diff
@@ -21,15 +22,20 @@ const Eval3 = {
...
-        "IBM-MIT-Harvard-Stanford": "mit_",
-        "OPICS (OSU, UU, NYU)": "opics_",
-        "MESS-UCBerkeley": "mess_",
-    }
+const Eval3_5 = {
+    moviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.5",
+    interactiveMoviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.5/",
+    topDownMoviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.5/",
+    movieExtension: ".mp4",
+    sceneBucket: "https://evaluation-images.s3.amazonaws.com/eval-scenes-3.5/",
```
brianpippin, line 32: Just so you know I had to name it differently this way because of collision on the lambda puts, eval-3.5-scenes wouldn't have worked, which is why it is this convention and should stay this way going forward.

analysis-ui/src/components/Analysis/evalConstants.jsx
rartiss55: Yup, I referenced the ticket in the comment above, and I added your comment in there about maybe using an admin page to manage it.

PR #25, opened 2021-03-16
MCS-587: Add a configurable Evaluation Status Page

brianpippin:
Show the progress of the evaluation

### DIFF
--- a/analysis-ui/src/components/EvaluationStatus/evalStatusPage.jsx
+++ b/analysis-ui/src/components/EvaluationStatus/evalStatusPage.jsx
@@ -1,0 +1,53 @@
+import React from 'react';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+import EvalStatusTable from './evalStatusTable';
+
+const historyFieldQueryName = "getHistorySceneFieldAggregation";
+
+const history_field_aggregation = gql`
+    query getHistorySceneFieldAggregation($fieldName: String!){
+        getHistorySceneFieldAggregation(fieldName: $fieldName) 
+  }`;
+
+class EvalStatusPage extends React.Component {
+
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            currentEval: "",
+        }
+
+        this.selectEvaluation = this.selectEvaluation.bind(this);
+    }
+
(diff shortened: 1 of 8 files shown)


analysis-ui/src/components/EvaluationStatus/evalStatusPage.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+        this.state = {
+            currentEval: "",
+        }
+
+        this.selectEvaluation = this.selectEvaluation.bind(this);
+    }
+
+    selectEvaluation(val) {
+        console.log(val);
```
rartiss55, line 26: We can probably remove this right?

node-graphql/server.schema.js
```diff
@@ -248,6 +250,61 @@ const mcsResolvers = {
...
+            const metadatas =  await mcsDB.db.collection('mcs_history').distinct("metadata").then(result => {return result});
+
+            for(let i=0; i < performers.length; i++) {
+                performersArray.push({label: performers[i], value: performers[i]});
+            }
+            for(let i=0; i < metadatas.length; i++) {
+                metadataArray.push({label: metadatas[i], value: metadatas[i]});
+            }
+
+            const regexObj = {$options: 'i', $regex: ".*" + args.eval.substring(0, 13) + ".*"};
```
rartiss55, line 270: So this will definitely work but its making me wonder if we should have an evalNum field or something where we don't have to search through strings and can just look for "2" or "3.5" or whatever.

rartiss55 (approved):
Looks good! I deleted some scenes locally and re-ingested them to see it in action. 

The only thing we might need to come back to post-3.5 (that I'm sure you considered) is maybe adding some context for different categories (like "please note that for eval x, we only ran such and such category at level2, so level1 will have no results/show no progress") or maybe even making the configuration so that we don't display all metadata levels for all tests if it doesn't apply.

analysis-ui/src/components/EvaluationStatus/evalStatusPage.jsx
brianpippin: That whole variable is OBE because I moved it when refactoring so I could remove all of this, thanks!

PR #27, opened 2021-04-21
MCS-517 - [Analysis UI Page] Move scene info/objects in scene details to a modal

rartiss55:
- A modal for scene details/download has been added to the score table.
- You can now select the score table row to select a scene, instead of using a button. 
- Some refactoring -- most notably score table and details are now separate components.

### DIFF
--- a/analysis-ui/src/components/Analysis/sceneDetails.jsx
+++ b/analysis-ui/src/components/Analysis/sceneDetails.jsx
@@ -66,8 +66,8 @@ function SceneDetailsModal({show, onHide, currentSceneNum, currentScene, constan
                             Scene {currentSceneNum} Details 
                         </Modal.Title>
 
-                        {currentScene !== undefined
-                            && currentScene !== null && <div className="download-scene">
+                        {currentScene !== undefined && currentScene !== null &&
+                        <div className="download-scene">
                             <a href={constantsObject["sceneBucket"] + currentScene.name + constantsObject["sceneExtension"]} download>
                                 <i className='material-icons'>get_app</i>
                                 <span className="download-scene-text">Download .JSON</span>
--- a/analysis-ui/src/components/Analysis/sceneDetails.jsx
+++ b/analysis-ui/src/components/Analysis/sceneDetails.jsx
@@ -1,0 +1,138 @@
+import React, {useState} from 'react';
+import $ from 'jquery';
+import Modal from 'react-bootstrap/Modal';
+import Button from 'react-bootstrap/Button'
+import { convertValueToString } from './displayTextUtils';
+
+function SceneDetailsModal({show, onHide, currentSceneNum, currentScene, constantsObject}) {
+    
+    const [currentObjectNum, setCurrentObjectNum] = useState(0);
+
+    const closeModal = () => {
+        onHide();
+    }
+
+    const checkSceneObjectKey = (scene, objectKey, key, labelPrefix = "") => {
+        if(objectKey !== 'objects' && objectKey !== 'goal' && objectKey !== 'name') {
+            return (
+                <tr key={'scene_prop_' + key}>
+                    <td className="bold-label">{labelPrefix + objectKey}:</td>
+                    <td className="scene-text">{convertValueToString(scene[objectKey])}</td>
+                </tr>
+            );
+        } else if(objectKey === 'goal') {
+            return (
+                Object.keys(scene["goal"]).map((goalObjectKey, goalKey) => 
+                    checkSceneObjectKey(scene["goal"], goalObjectKey, goalKey, "goal."))
+            );
+        } else if(objectKey === 'name') {
+            return (
+                <tr key={'scene_prop_' + key}>
+                    <td className="bold-label">{labelPrefix + objectKey}:</td>
+                    <td className="scene-text">{convertValueToString(scene[objectKey])} (<a href={constantsObject["sceneBucket"] + scene[objectKey] + constantsObject["sceneExtension"]} download>Download Scene File</a>)</td>
+                </tr>
+            );
+        }
+    }
+
+    const findObjectTabName = (sceneObject) => {
+        if(sceneObject.shape !== undefined && sceneObject.shape !== null) {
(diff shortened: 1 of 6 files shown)


analysis-ui/src/components/Analysis/sceneDetails.jsx
```diff
@@ -0,0 +1,160 @@
...
+            return "occluder wall";
+        }
+
+        if(sceneObject.id.indexOf('occluder_pole') > -1) {
+            return "occluder pole";
+        }
+
+        if(sceneObject.role !== undefined && sceneObject.role !== null) {
+            return sceneObject.role;
+        }
```
rartiss55, line 48: Also fixed some bugs here with displaying tab titles properly for "Objects in Scene" + info wasn't correct for agency.

analysis-ui/src/components/Analysis/scenes.jsx
```diff
@@ -54,14 +56,22 @@ const mcs_scene= gql`
...
+]
 class Scenes extends React.Component {
 
     constructor(props) {
         super(props);
         this.state = {
             currentPerformerKey: 0,
             currentPerformer: props.value.performer !== undefined ? props.value.performer : "",
-            currentSceneNum: (props.value.scene !== undefined && props.value.scene !== null) ? parseInt(props.value.scene) - 1 : 0,
+            currentSceneNum: (props.value.scene !== undefined && props.value.scene !== null) ? parseInt(props.value.scene) : 1,
```
rartiss55, line 74: Had to fix this for eval 2 page -- indexing was a bit wonky since for eval 3, we use a map instead of an array to track scene history files, so the new components were expecting a 1 instead of a 0.

analysis-ui/src/components/Analysis/sceneDetails.jsx
```diff
@@ -0,0 +1,160 @@
...
+                        {currentScene !== undefined && currentScene !== null &&
+                        <div className="download-scene">
+                            <a href={constantsObject["sceneBucket"] + currentScene.name + constantsObject["sceneExtension"]} download>
+                                <i className='material-icons'>get_app</i>
+                                <span className="download-scene-text">Download .JSON</span>
+                            </a>
+                        </div>}
+                    </div>
+
+                    <div className="scene-details-btn-group btn-group" role="group">
```
rartiss55, line 78: Mentioned this a bit at standup: I removed the goal.object_list from the "Scene Details" since it seemed redundant (we have "Objects in Scene" for that) but wondering if there were other ways to improve this section. I think we agreed to just leave everything there "just in case", but wondering if we should delete any of the really hard to read sections (i.e. goal.sceneInfo) since people can always download the JSON file/that might actually be easier to read? 

I had experimented a bit with removing fields that were false/empty, but it was still pretty hard to read.

rartiss55:
Found a bug with this -- I'll close and reopen when fixed.

rartiss55:
Turns out the problem I found earlier was with duplicate mongo records and nothing specific to this PR (hope I didn't give anyone whiplash with my false alarm) -- reopening :)

analysis-ui/src/components/Analysis/sceneDetails.jsx
```diff
@@ -0,0 +1,160 @@
...
+import $ from 'jquery';
+import Modal from 'react-bootstrap/Modal';
+import Button from 'react-bootstrap/Button'
+import { convertValueToString } from './displayTextUtils';
+
+function SceneDetailsModal({show, onHide, currentSceneNum, currentScene, constantsObject}) {
+    
+    const [currentObjectNum, setCurrentObjectNum] = useState(0);
+    const [selectedData, setSelectedData] = useState("scene_info");
+    const keysToIgnore = ['objects', 'goal', 'objectsInfo'];
```
brianpippin, line 11: I am thinking we should explore if this would be better to be implemented on the server side.  I think it would probably speed up page load, and we shouldn't return extra information from the server we don't plan to use.

brianpippin: Nevermind, I think I implemented this, and we are just excluding the objects to be rendered elsewhere, feel free to ignore the above.

analysis-ui/src/components/Analysis/sceneDetails.jsx
brianpippin: I think we might just need to totally clean up the code to potentially render, exactly what we want and some point and not make it so dynamic?  I think at this point we know what fields are in the scenes, we could have something that is a list of fields we want to render and different ways to render based on the object type?  Perhaps a future improvement?

brianpippin:
I got the following error when sorting on the second column and opening details: 

TypeError: Cannot convert undefined or null to object
SceneDetailsModal
src/app/src/components/Analysis/sceneDetails.jsx:116
  113 | </div>
  114 | <div className="object-contents">
  115 |     <table>
> 116 |         <tbody>
      | ^  117 |             { Object.keys(currentScene.objects[currentObjectNum]).map((objectKey, key) => 
  118 |                 <tr key={'object_tab_' + key}>
  119 |                     <td className="bold-label">{objectKey}:</td>

analysis-ui/src/components/Analysis/sceneDetails.jsx
rartiss55: I put together a ticket here: https://nextcentury.atlassian.net/browse/MCS-645

rartiss55:
> I got the following error when sorting on the second column and opening details:
> 
> TypeError: Cannot convert undefined or null to object
> SceneDetailsModal
> src/app/src/components/Analysis/sceneDetails.jsx:116
> 113 |
> 114 |
> 
> 
> 115 |
> > 116 |
> 
> 
> | ^ 117 | { Object.keys(currentScene.objects[currentObjectNum]).map((objectKey, key) =>
> 118 | <tr key={'object_tab_' + key}>
> 119 |
> {objectKey}:

Thanks for finding this! I fixed this by resetting things when we close the modal (the current object index + the current tab).

brianpippin (approved):

PR #28, opened 2021-04-22
MCS-583: Remove usage of old keys collection from codebase

brianpippin:
This has a companion ticket in mcs-ingest that actually drops the collection from the database, along with setting up the framework for deployment scripts to be run.


ThomasSchellenbergNextCentury (approved):
Code check seems fine.

rartiss55 (approved):

PR #29, opened 2021-04-28
Refactoring Analysis UI Page(s)

rartiss55:
- Renaming queries on analysis view for more clarity going forward. 
- Similarly, renamed scenes.jsx to scenesEval2.jsx, and scenesEval3.jsx is now scenes.jsx (this makes the overall diff a bit weird, might be easier to view commits individually).
- Split out classification by step table and interactive scene player into separate components to make scenes.jsx less painful to look at. 
- Switched to using React hooks and utilizing the className attribute instead of document.getElementById/jquery for interactive scene player component.


analysis-ui/src/components/App/index.jsx
```diff
@@ -62,7 +62,7 @@ const AnalysisUI = ({newState, updateHandler}) => {
     }
 
     let hasEval =  (newState.eval !== undefined && newState.eval !== null)
-    let isEval3 = hasEval && newState.eval.includes("3");
+    let isEval2 = hasEval && newState.eval.includes("2");
```
brianpippin, line 65: We might want to make this includes(" 2 ") just in case we ever do something like eval "4.25" or something dumb like that lol

brianpippin (approved):
I tried to break the UI, I could not, so it seems like it is good to go :)

analysis-ui/src/components/App/index.jsx
rartiss55: Good point! Done.

Thanks for pointing this out cuz it also helped me notice a couple more small things we might want to update with how the queries are run and how the navigation logic works (even though I'm hoping I get to redo that last bit before Eval 4). Leaving this note for myself for Monday.

analysis-ui/src/components/Analysis/classificationByStepTable.jsx
```diff
@@ -0,0 +1,104 @@
...
+import _ from "lodash";
+import { convertValueToString } from './displayTextUtils';
+
+function ClassificationByStepTable ({evaluation, currentSceneHistItem}) {
+
+    const [showInternalState, toggleShowInternalState] = useState(false);
+
+    const showInternalStatePreview = (stepObj) => {
+        if(!_.isEmpty(stepObj.internal_state)) {
+            return JSON.stringify(stepObj.internal_state).substring(0, 20) + '...';
```
ThomasSchellenbergNextCentury, line 13: I'm assuming we're showing just the first 20 characters here because it's a preview? Should `20` be its own variable?

analysis-ui/src/components/Analysis/classificationByStepTable.jsx
```diff
@@ -0,0 +1,104 @@
...
+
+        } else {
+            return JSON.stringify(stepObj.internal_state);
```
ThomasSchellenbergNextCentury, line 16: Isn't the `internal_state` empty in this condition? If so, could you just return `"{}"` here?

analysis-ui/src/components/Analysis/interactiveScenePlayer.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+
+    const goToVideoLocation = (location) => {
+        if( document.getElementById("interactiveMoviePlayer") !== null && location !== currentStep) {
+            if(evaluation === "Evaluation 2 Results") {
+                setCurrentStep(location);
+                scenePlayer.current.currentTime = location;
+            } else {
+                // videos for eval 3+ are faster than eval 2 -- 20 actions/frames per second
+                let timeToJumpTo = (location + 1) / 20;
```
ThomasSchellenbergNextCentury, line 25: We may want to make a global variable for the `20` FPS since it may be configurable in the future (so it would be easier to find all the places we need to update).

analysis-ui/src/components/Analysis/interactiveScenePlayer.jsx
```diff
@@ -0,0 +1,97 @@
...
+                stepElems.current[stepIndex].scrollIntoView({behavior: "smooth", block: "nearest", inline: "start"});
+            }
+        }
+    }
+
+    return (
+        <div className="movie-steps-holder">
+            <div className="interactive-movie-holder">
+                <video id="interactiveMoviePlayer" ref={scenePlayer}
+                src={sceneVidLink} width="500" height="350" controls="controls" autoPlay={false} onTimeUpdate={highlightStep} onLoadedData={initializeStepView}/>
```
ThomasSchellenbergNextCentury, line 75: Why are the width and height 500x350 here? Shouldn't it be 600x400? (Same with the top-down video.)

analysis-ui/src/components/Analysis/scenes.jsx
```diff
@@ -2,41 +2,49 @@ import React from 'react';
 import { Query } from 'react-apollo';
 import gql from 'graphql-tag';
 import _ from "lodash";
-import $ from 'jquery';
-import FlagCheckboxMutation from './flagCheckboxMutation';
+//import FlagCheckboxMutation from './flagCheckboxMutation';
```
ThomasSchellenbergNextCentury, line 5: Does this line need to be kept for any reason?

analysis-ui/src/components/Analysis/scenes.jsx
```diff
@@ -2,41 +2,49 @@ import React from 'react';
...
 
-const historyQueryName = "getEvalHistory";
+const historyQueryName = "createComplexQuery";
+const historyQueryResults = "results";
 const sceneQueryName = "getEvalScene";
 
 let constantsObject = {};
-let currentState = {};
-let currentStep = 0;
+//let currentState = {};
```
ThomasSchellenbergNextCentury, line 16: Same with this line.

analysis-ui/src/components/Analysis/scenes.jsx
```diff
@@ -59,44 +65,55 @@ const setConstants = function(evalNum) {
...
         this.state = {
-            currentPerformerKey: 0,
             currentPerformer: props.value.performer !== undefined ? props.value.performer : "",
+            currentMetadataLevel: props.value.metadata_lvl !== undefined ? props.value.metadata_lvl : "",
             currentSceneNum: (props.value.scene !== undefined && props.value.scene !== null) ? parseInt(props.value.scene) : 1,
-            currentObjectNum: 0,
-            flagRemove: false,
-            flagInterest: false,
+            //flagRemove: false,
+            //flagInterest: false,
```
ThomasSchellenbergNextCentury, line 84: Same with these lines.

analysis-ui/src/components/Analysis/scenes.jsx
```diff
@@ -59,44 +65,55 @@ const setConstants = function(evalNum) {
...
         }
 
+        /*
         if(this.state.currentPerformer === "" && firstEval !== null && firstEval !== undefined) {
             this.setState({
                 flagRemove: firstEval["flags"]["remove"],
                 flagInterest: firstEval["flags"]["interest"]
             });
-        }
+        }*/
```
ThomasSchellenbergNextCentury, line 112: Same with this block.

rartiss55: I think at the time this was functionality we were planning to fix later. I believe we might have decided that this will be taken out after all, but I left it in since this that's likely covered by another ticket.

analysis-ui/src/components/Analysis/interactiveScenePlayer.jsx
rartiss55: I think this was more about screen real estate, since this video component has a scrollable action list in the middle of two videos. We have tickets to revisit the video components later as well, this was just a first stab refactor to make doing that a bit easier.

analysis-ui/src/components/Analysis/scenes.jsx
rartiss55: See comment below for answer to these.

analysis-ui/src/components/Analysis/scenes.jsx
rartiss55: Looking at this again, also want to clarify that this wasn't just commented out recently, its just that the diff is weird with the file renaming.

analysis-ui/src/components/Analysis/classificationByStepTable.jsx
rartiss55: You're right! Returning an empty string here now.

analysis-ui/src/components/Analysis/interactiveScenePlayer.jsx
rartiss55: Made a variable for this.

Note though that if/when this changes, it'll probably have to be an additional variable and additional logic for backwards compatibility purposes.

ThomasSchellenbergNextCentury: Makes sense. Thanks.

analysis-ui/src/components/Analysis/interactiveScenePlayer.jsx
ThomasSchellenbergNextCentury: Sounds good.

analysis-ui/src/components/Analysis/scenes.jsx
ThomasSchellenbergNextCentury: Ah, OK, that makes more sense. Might still be helpful to have a few TODO comments here though.

rartiss55: Added a TODO.

analysis-ui/src/components/Analysis/classificationByStepTable.jsx
rartiss55: I didn't end up making it its own variable but at least left a comment for clarity.

ThomasSchellenbergNextCentury: Awesome. Thanks.

PR #30, opened 2021-04-28
MCS-519: remove comments component and MCS-589: update eval constants 

rartiss55:
Was going to tack this onto the refactor branch, but separated it since that diff is rather large already.


analysis-ui/src/components/Analysis/evalConstants.jsx
```diff
@@ -33,9 +33,58 @@ const Eval3_5 = {
...
 export const EvalConstants = {
     "Evaluation 2 Results": Eval2, 
     "Evaluation 3 Results": Eval3,
-    "Evaluation 3.5 Results": Eval3_5
+    "Evaluation 3.5 Results": Eval3_5,
+    "Evaluation 4 Results": Eval4,
+    "Evaluation 5 Results": Eval5,
+    "Evaluation 6 Results": Eval6,
+    "Evaluation 7 Results": Eval7,
+    "Evaluation 8 Results": Eval8
```
rartiss55, line 89: Dunno if there will be an Eval 8 now, but just made this consistent with what's in ingest. Made an additional ticket to update this + ingest in case we need to rerun anything before Eval 4 (https://nextcentury.atlassian.net/browse/MCS-646).

brianpippin (approved):

analysis-ui/src/components/Analysis/evalConstants.jsx
```diff
@@ -33,9 +33,58 @@ const Eval3_5 = {
...
 
-// TODO: MCS-589: move to ingest/store in mongo
+const Eval4 = {
+    moviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-4/",
+    interactiveMoviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-4/",
+    topDownMoviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-4/",
+    movieExtension: ".mp4",
+    sceneBucket: "https://evaluation-images.s3.amazonaws.com/eval-4-scenes/",
+    sceneExtension: "_debug.json",
+}
```
ThomasSchellenbergNextCentury, line 43: If there's a consistent pattern to these variables, and we don't expect it to change in the future, could we just write a function that takes the eval number and returns the corresponding dict, and use the function to set each variable?

rartiss55: I like the idea, but there is a possibility it could change in the future, so I'm leaning towards leaving it as is.

ThomasSchellenbergNextCentury: Sounds good!

ThomasSchellenbergNextCentury (approved):

PR #31, opened 2021-05-04
Fixes for evaluation references on homepage/header drop downs

rartiss55:
- Fixing places where we were checking for eval 2 code with "includes("2")" (I used that when we weren't really sure about naming conventions, but now we can reliably use "Evaluation 2 Results")
- Likewise, removing anything specifically looking for eval 3 like "includes("3")" from the header dropdown on the Analysis UI page so that future evals like Eval 4 still work. 
- Adding eval argument to the history/scene queries, so we're not just depending on category type/test type and number
- Fixed a bug with eval regex in server.schema

### DIFF
--- a/analysis-ui/src/components/Analysis/header.jsx
+++ b/analysis-ui/src/components/Analysis/header.jsx
@@ -71,7 +71,7 @@ class ListItem extends React.Component {
             if(this.props.stateName === 'eval') {
                 // We have 2 cases: "eval 2" and "every other eval"
                 let isValidEval2 = hasTestTypeState && this.props.item === EVAL_2_IDENTIFIER;
-                let isValidOtherEval = hasCatTypeState && (this.props.item !== EVAL_2_IDENTIFIER);
+                let isValidOtherEval = hasCatTypeState && this.props.item !== EVAL_2_IDENTIFIER;
 
                 if(isValidEval2) {
                     paramsToAppend += "&test_type=" + this.props.state["test_type"];
@@ -208,7 +208,7 @@ class EvalNav extends React.Component {
                         <DropListItems fieldName={"eval"} stateName={"eval"} state={this.props.state} updateHandler={this.props.updateHandler}/>
                     </NavDropdown>
 
-                    {(this.props.state.eval !== undefined && this.props.state.eval !== null && (this.props.state.eval !== EVAL_2_IDENTIFIER)) &&
+                    {(this.props.state.eval !== undefined && this.props.state.eval !== null && this.props.state.eval !== EVAL_2_IDENTIFIER) &&
                     <NavDropdown title={"Category Type: " + (
                         (this.props.state.category_type === undefined || this.props.state.category_type === null) ? "None" : this.props.state.category_type)}
                         id="basic-nav-dropdown">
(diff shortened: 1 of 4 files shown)


node-graphql/server.schema.js
```diff
@@ -131,6 +131,10 @@ const mcsTypeDefs = gql`
   }
 `;
 
+const getEvalNumFromString = (strValue) => {
+    return strValue.replace(/\D+\.?\D+/g, "");
```
rartiss55, line 135: Updated this check from getComplexResults since it wasn't accounting for decimals ("3.5" would become "35").

analysis-ui/src/components/Analysis/header.jsx
```diff
@@ -31,6 +31,8 @@ const GET_SUBMISSION_AGG = gql`
         }
   }`;
 
+const EVAL_2_IDENTIFIER = "Evaluation 2 Results";
+
 class ListItem extends React.Component {
```
rartiss55, line 36: There is an additional ticket to redo this navigation logic -- this is just so the current version isn't broken in case we don't get to it before eval 4.

analysis-ui/src/components/Analysis/header.jsx
```diff
@@ -59,29 +61,30 @@ class ListItem extends React.Component {
...
 
                 if(isValidEval2) {
                     paramsToAppend += "&test_type=" + this.props.state["test_type"];
-                } else if(isValidEval3) {
+                } else if(isValidOtherEval) {
                     paramsToAppend += "&category_type=" + this.props.state["category_type"];
                 }
 
-                if(hasTestNumState && (isValidEval2 || isValidEval3)) {
+                if(hasTestNumState && (isValidEval2 || isValidOtherEval)) {
```
brianpippin, line 82: At this point with all our refactoring is it every possible for one of those to be False?  It would seem like we wouldn't need to have this check because one of them should be True, and we could just check hasTestNumState, although I might be forgetting a scenario

analysis-ui/src/components/Analysis/header.jsx
```diff
@@ -205,14 +208,14 @@ class EvalNav extends React.Component {
                         <DropListItems fieldName={"eval"} stateName={"eval"} state={this.props.state} updateHandler={this.props.updateHandler}/>
                     </NavDropdown>
 
-                    {(this.props.state.eval !== undefined && this.props.state.eval !== null && this.props.state.eval.includes("3")) &&
+                    {(this.props.state.eval !== undefined && this.props.state.eval !== null && (this.props.state.eval !== EVAL_2_IDENTIFIER)) &&
```
brianpippin, line 211: Super super nitpick, but not sure you need the extra parens you added.

rartiss55: you're right, took this and another set out

analysis-ui/src/components/Analysis/header.jsx
rartiss55: Tested it out again -- looks like this logic ensures that the test_num is cleared if you go back and select a new Eval, so I think it should stay in.

brianpippin (approved):

PR #33, opened 2021-05-26
Constants update for gravity rerun (MCS-646).

rartiss55:
Still need to update triggers for automated ingest (see comment).

### DIFF
--- a/analysis-ui/src/components/Analysis/evalConstants.jsx
+++ b/analysis-ui/src/components/Analysis/evalConstants.jsx
@@ -33,6 +33,15 @@ const Eval3_5 = {
     sceneExtension: "_debug.json",
 }
 
+const Eval3_75 = {
+    moviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.75/",
+    interactiveMoviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.75/",
+    topDownMoviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.75/",
+    movieExtension: ".mp4",
+    sceneBucket: "https://evaluation-images.s3.amazonaws.com/eval-scenes-3.5/", 
+    sceneExtension: "_debug.json",
+}
+
 const Eval4 = {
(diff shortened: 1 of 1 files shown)


analysis-ui/src/components/Analysis/evalConstants.jsx
```diff
@@ -33,6 +33,15 @@ const Eval3_5 = {
     sceneExtension: "_debug.json",
 }
 
+const Eval3_75 = {
+    moviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.75/",
+    interactiveMoviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.75/",
+    topDownMoviesBucket: "https://evaluation-images.s3.amazonaws.com/eval-3.75/",
+    movieExtension: ".mp4",
+    sceneBucket: "https://evaluation-images.s3.amazonaws.com/eval-scenes-3.5/",
```
rartiss55, line 41: I only made a new folder for the history files, but not the scene bucket since I wasn't sure about this: I thiiiiink we wanted to reuse the same scenes we had previously, but maybe we need to copy them over again to a new folder for automated ingest? 

I can update the triggers to whatever they need to be once this is clarified.

brianpippin: I think we want to copy the scenes into a new bucket, I don't think we have code to have a mismatch on the scenes/history files evaluation names.  The only problem is if those scene files actually have evaluation-3.5 in them, then we would need to modify every scene file we copied.  It should be easy enough with a script but that might be work we need to figure out.  As for this code though, I believe we want to update the sceneBucket.

ThomasSchellenbergNextCentury (approved):
The code looks good.

analysis-ui/src/components/Analysis/evalConstants.jsx
rartiss55: Okay, I made a ticket here to copy over things correctly for 3.75 (https://nextcentury.atlassian.net/browse/MCS-690), and made another folder for the 3.75 scene files/updated the constants.

brianpippin (approved):

PR #34, opened 2021-05-27
Update urls to use domains

brianpippin:
MCS-617 & 605:  Code to allow for us to use subdomain urls and add https functionality.


rartiss55 (approved):

ThomasSchellenbergNextCentury (approved):
Looking a lot nicer!

PR #35, opened 2021-06-24
MCS-615: Update bucket URL 

brianpippin:
Putting S3 behind CloudFront/Route 53 DNS/subdomain, also needed for MediaConvert Lambda function.


ThomasSchellenbergNextCentury (approved):
Code LGTM

rartiss55:
Looks good. I noticed an old s3 link reference in results.jsx (line 10), but is that legacy code?

brianpippin:
> Looks good. I noticed an old s3 link reference in results.jsx (line 10), but is that legacy code?

Yes that link is for the intphys images, probably from Eval 1

PR #36, opened 2021-06-30
Fixed query table header

jaudick:


rartiss55:
Works, but something about the stickyHeader property made the other cell borders in the table bolder? At a quick glance looks like it has a different border-collapse property, so we probably just need to override that. 

Also, it'd be nice if the results per page selection was underneath the scrollable section rather than within in, so one could  adjust that again without scrolling through a bunch of results.

brianpippin:
I pushed some fixes to this branch to get this the rest of the way.  Rachel - I needed to make the header a different color so it would be set off when scrolling, this how all the examples do the sticky header, you can't do it with just a border, it looks weird.  Let me know if this all works for you both.

jaudick:
@brianpippin The new changes look great to me

rartiss55:
Oh my comment was actually about extra cell borders in the rest of the table, not the header, but those are gone now, so LGTM!

rartiss55 (approved):

brianpippin:
> Oh my comment was actually about extra cell borders in the rest of the table, not the header, but those are gone now, so LGTM!

I saw those lines as well, don't know what the sticky-header added but this css fixes it:  border-collapse: collapse;

brianpippin (approved):

PR #37, opened 2021-07-09
Variable playback rate buttons on analysis page movies

jaudick:


rartiss55:
@jaudick We did have some old mockups of how we wanted the play buttons to look (they don't include the syncing, but it might still be helpful). I invited you to the page (tagging @brianpippin in case his guidance said to ignore these).

analysis-ui/src/components/Analysis/playbackButtons.jsx
```diff
@@ -0,0 +1,94 @@
...
+const topDownMovie = "topDownInteractiveMoviePlayer";
+
+class PlaybackButtons extends React.Component{
+
+    constructor(props) {
+        super(props);
+    }
+
+    setSpeed = (speed, btnName, className, movieId) => {
+        document.getElementById(movieId).playbackRate = speed;
```
rartiss55, line 17: document.getElementId should work in these cases, but I think in react its recommended to use ref when you can. Just a heads up.

analysis-ui/src/components/Analysis/playbackButtons.jsx
```diff
@@ -0,0 +1,94 @@
...
+        this.setSpeed(1, "1x-btn", "scene-movie-playback-button", sceneMovie);
+        this.setSpeed(1, "td-1x-btn", "topdown-movie-playback-button", topDownMovie);
+        this.setLoop("scene-loop-video-button",sceneMovie, true);
+        this.setLoop("topdown-loop-video-button",topDownMovie, true);
+    }
+
+    render() {
+        return (
+            <div id="playbackButtons" className="movie-playback-buttons-holder">
+                <div className="scene-movie-playback-buttons-holder">
```
rartiss55, line 51: Each group of playback buttons could be a component so you'd have less repeated code here.

rartiss55:
I tested it out, works well -- the only thing I noticed is that you may want to sync the playback speed for all videos if they are different and you then click 'sync all'.

brianpippin:
Jacob the functionality is great, I would just like to see some UI clean up.  I think the variable speeds should be a dropdown, and it should apply to both movies so that if you do .5 it applies to both movies.  I also think the loop button should apply to both as well, so you're looping through them, and I think the play button should just apply to both.  I can't imagine a scenario where it wouldn't hurt to have both movies playing at the same time, so never let them play one separately over the other that way we only have one set of controls in the middle.  Then if you could, see if you can make the buttons look closer to the mockups.  

Screen Shot 2021-07-16 at 2 51 04 PM

brianpippin:
Ignore the comment about playing separately, i think we still want the movie controls that come with each player, so the sync play button should still say that, with the loop button next to it applying to both, and a dropdown with speeds on the other side of the sync button, i forsee them using those button in place of the player controls but we want to leave the player controls I think.

jaudick:
@brianpippin These comments make sense. I will get on it!

jaudick:
@brianpippin @rartiss55 new ui updated. I made sure to test every combination of switching between scenes and evals and tests. Hitting "Play all" does not play all the movies with the click, rather it enables the movies to go to the next one after the movie ended and then  autoplay the next movie. I added a pause even though its not in the ui mockup. Hitting play while a movie is playing resets it to the beginning.

rartiss55 (approved):

analysis-ui/src/components/Analysis/interactiveScenePlayer.jsx
```diff
@@ -70,29 +72,32 @@ function InteractiveScenePlayer({evaluation, sceneVidLink, topDownLink, sceneHis
...
-                        sceneHistoryItem.steps.map((stepObject, key) => 
-                        <div key={"step_div_" + key} id={"stepHolder" + (key+1)} ref = {element => stepElems.current[key] = element} className={currentStep === key+1 ? "step-div step-highlight" : "step-div"} onClick={() => goToVideoLocation(key+1)}>
-                            {stepObject.stepNumber + ": " + stepObject.action + " (" + convertValueToString(stepObject.args) + ") - " + stepObject.output.return_status}
-                        </div>
-                    )}
+        <div>
+            <div className="movie-steps-holder">
+                <div className="interactive-movie-holder">
+                    <video id="interactiveMoviePlayer" ref={scenePlayer}
+                    src={sceneVidLink} width="500" height="350" controls="controls" onTimeUpdate={highlightStep} onLoadedData={initializeStepView}
```
brianpippin, line 79: We might want to put the width and height of the video players in a constant, so it's easier to change them in the future, especially if we want to maybe dynamically inc/dec size of the player.

brianpippin:
Screen Shot 2021-07-26 at 2 19 02 PM
So code looks great and the behavior is amazing, I have a few nitpicks, sorry!  The controls should be centered underneath both videos if possible.  I am not sure what the play all button does that the play button doesn't do.  I think it would actually be cool if the play button either changed to the pause button or to a stop button, right now you need to click one or the other.  Also the margin above the controls is not the same as the margin below the controls, so you might need some margin or padding underneath.  Everything is looking great and is super functional, just trying to polish it off!

rartiss55:
> I am not sure what the play all button does that the play button doesn't do.

It goes through the whole list of scenes as opposed to playing videos for a single scene.

brianpippin:
@jaudick - I don't know if you saw my previous comments or forgot to push an update.

jaudick:
@brianpippin location fixed

PR #38, opened 2021-07-12
Color of every performer is displayed throughout all ui pages

jaudick:
I got the rgb values from the css displayed in the web console

### DIFF
--- a/analysis-ui/src/components/Analysis/scenes.jsx
+++ b/analysis-ui/src/components/Analysis/scenes.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { Query } from 'react-apollo';
 import gql from 'graphql-tag';
-import _ from "lodash";
+import _, { get } from "lodash";
 // TODO: Remove FlagCheckboxMutation component + currentState + related variables?
 //import FlagCheckboxMutation from './flagCheckboxMutation';
 import {EvalConstants} from './evalConstants';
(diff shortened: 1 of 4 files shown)


analysis-ui/src/components/Analysis/scenes.jsx
```diff
@@ -192,6 +192,33 @@ class Scenes extends React.Component {
...
+        this.setState({[key]: value}, () => {
+            colorCallback(value, id);
+        });
+    }
+    
+    getStyleForPerformer(performer, id="") {
+        const getColor = (performer) => {
+            return (
+                performer === this.state.currentPerformer ? '#0062cc' :
+                String(performer).includes("IBM-MIT-Harvard-Stanford") ? "rgb(0, 160, 210, 1)" :
```
rartiss55, line 205: Constants would be good for these performer names (sounds like the one for MIT/CORA is going to be updated, so it would be good just to have to change it in one place).

analysis-ui/src/components/Analysis/scenes.jsx
```diff
@@ -192,6 +192,33 @@ class Scenes extends React.Component {
...
+                String(performer).includes("MESS-UCBerkeley") ? "rgb(68, 77, 93, 1)" :
+                String(performer).includes("OPICS") ? "rgb(217, 85, 85, 1)" :
```
rartiss55, line 207: I get why this is 'includes' instead of an equality check, since there are a few performers with things like "2" or "Rerun" appended in special cases, but it should be safe to say the OPICS base name is "OPICS (OSU, UU, NYU)"

rartiss55:
With this ticket, would it be good to include the home page/resultsChart and update those colors? (more posing the question -- it might not be possible since that may be managed by an outside library)

analysis-ui/src/components/Analysis/scenes.jsx
```diff
@@ -192,6 +192,33 @@ class Scenes extends React.Component {
...
+                String(performer).includes("TA2 Baseline") ? "rgb(138, 85, 217, 1)" :
+                "inherit"
+            );
+        }
+
+        if(id!=="") {
```
rartiss55, line 213: Couldn't you just return 'getColor(performer)' without doing the document.getElementById check, since its being called from that element already in the style tag?

brianpippin:
> With this ticket, would it be good to include the home page/resultsChart and update those colors? (more posing the question -- it might not be possible since that may be managed by an outside library)

So I might not have been specific enough on this, but this is primarily targeted at the homepage result charts, so that every performer is the same there and then carry that color through all the charts that may follow like with plausibility.  I haven't loaded this up yet to see where the colors were added specifically but the results chart was the primary place this should be occurring.

jaudick:
@brianpippin Oh my I did an entirely different task then. I based all the colors on the eval 3.5 results charts and spread those colors consistently out over all the pages. There may be value in the new aesthetic but now hearing what the primary target was, what I did might not make any sense when looking at the other eval charts

rartiss55:
@jaudick @brianpippin did you hash out the original intention of this ticket vs changes made in this PR?

brianpippin:
> @jaudick @brianpippin did you hash out the original intention of this ticket vs changes made in this PR?

I have this open was about to look at it, until I needed to figure out Lisa's CSV issues, will be looking at it soon.

brianpippin:
After looking this over and looking the ticket over, I unfortunately thing we need to scrap this branch.  The main idea of the original ticket was the coloring of the bar charts and the coloring of plausibility charts, nothing to do with the tables.  But now that we have the pipeline and have mess1 and mess2 we need to come up with a intuitive way to handle multiple submissions for a performer but keep the color schemes for all of these potential submissions in unison.

PR #40, opened 2021-07-16
MCS-764:  Update Eval Status Page every second automatically

brianpippin:
The eval status page will update every second if you are on the page so it will constantly rerender the updates on our process so no need to have a button as well.

### DIFF
--- a/analysis-ui/src/components/EvaluationStatus/evalStatusPage.jsx
+++ b/analysis-ui/src/components/EvaluationStatus/evalStatusPage.jsx
@@ -14,6 +14,22 @@ class EvalStatusPage extends React.Component {
 
     constructor(props) {
         super(props);
+        this.state = {
+            counter: 0
+        }
+        this.incrementCounter = this.incrementCounter.bind(this);
+    }
+
+    incrementCounter() {
+        this.setState({counter: this.state.counter + 1});
+    }
+
+    componentDidMount() {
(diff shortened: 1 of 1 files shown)


analysis-ui/src/components/EvaluationStatus/evalStatusPage.jsx
```diff
@@ -14,6 +14,22 @@ class EvalStatusPage extends React.Component {
 
     constructor(props) {
         super(props);
+        this.state = {
+            counter: 0
+        }
+        this.incrementCounter = this.incrementCounter.bind(this);
```
ThomasSchellenbergNextCentury, line 20: Out of curiosity, is there any specific benefit to using `bind` here, rather than on line 28, like `setInterval(this.incrementCounter.bind(this), 1000)`? Or is it just personal preference?

brianpippin: It is just a habit because I was thinking I was going to potentially be passing that function into another component when using the button to refresh, in which case you would need to use bind if you were passing that function around, but because it is only called inside this component it is probably redundant.

deanwetherby (approved):

ThomasSchellenbergNextCentury (approved):

jaudick (approved):

PR #41, opened 2021-07-19
Name suggestion based on Query Parameters for save query results

jaudick:
There is a "Combine All" option to combine all the query parameters across the different queries. The z-index style and css i added because the sticky header for the table seems to overlap the dropdowns for the query searches. I felt that was a bug fix somewhat related to the task so I fixed it. The only way to fix that z-overlap in some cases was the apply it directly to the component.

### DIFF
--- a/analysis-ui/src/components/QueryBuilder/fieldDropdown.jsx
+++ b/analysis-ui/src/components/QueryBuilder/fieldDropdown.jsx
@@ -63,7 +63,7 @@ class FieldDropdown extends React.Component {
 
     render() {
         return (
-            <div className="query-field-selector">
+            <div style={{zIndex:'11'}} className="query-field-selector">
                 <FieldSelector fieldType={this.props.fieldType} onSelectHandler={this.props.selectFieldHandler}/>
             </div>
         );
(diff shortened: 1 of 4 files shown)


rartiss55 (approved):

analysis-ui/src/components/QueryBuilder/fieldDropdown.jsx
```diff
@@ -63,7 +63,7 @@ class FieldDropdown extends React.Component {
 
     render() {
         return (
-            <div className="query-field-selector">
+            <div style={{zIndex:'11'}} className="query-field-selector">
```
brianpippin, line 66: I saw your comments.  You tried adding the z-index into the css for this class name and it didn't work?  I would really like to find a way to move the zindex out of the html into the css, did you also try "z-index: 11 !important" as well in that class?  Otherwise this looks and and works really well.

brianpippin (commented):
Check to see if you can get the z index out of the html after you do that, it would be good to merge.

jaudick:
@brianpippin z-index is now out of html with latest commit

analysis-ui/src/components/QueryBuilder/fieldDropdown.jsx
brianpippin: @jaudick I think you missed this one z-index, if you can add a class or to the class, then you can merge.

brianpippin (approved):

PR #44, opened 2021-07-21
Fix debug scene download link for Eval 3.75 scenes.

rartiss55:


brianpippin:
Good find, looks good!

PR #45, opened 2021-07-28
MCS-831: Fix for memory allocation error on prod

brianpippin:
Upgrade to Node 12 and increase swap space.


rartiss55 (approved):
Wondered if a node upgrade was needed and looks like it was :)

Rebuilt my node-server container and looks good -- I saw some warnings about deprecated dependencies (its been so long since I've built, maybe they've been there for a while or I just don't remember them), so I made minor priority ticket for those (MCS-835).

PR #46, opened 2021-08-03
MCS-467: Dynamically Create and Show CSV, along with adding Admin Features

brianpippin:
Features:
Test two things first:
1.  You should NOT have an Administrator link in the User Drop Down on the Right side of the NavBar
2.  You should NOT have a "Create CSV" button on the Evaluation Status Page

Now run the script that is in mcs-ingest to create the initial admin users, you should be one of them.
1.  Now you should be able to see the Adminstrator Link and the Create CSV link
2. You should be able to open the Create CSV Modal and select if you want to create a Scene or Results CSV
3. The CSV automatically get uploaded to the evaluation-images bucket (example link:  https://resources.machinecommonsense.com/csv-db-files/Evaluation_3.75_Results.csv)
4. If you want to double check the files are being created (as I have created most at this point you can delete it from the bucket or check the time stamp in the bucket)
5. The links to download the CSV appear after a pre-flight test to make sure those links exist, if they exist they show up, if they 404 they do not show up.
6. There is now an Adminstrator page under the right nav drop down.  Here only admin users can add or remove people to be admin users.  Currently Admin users can only create CSV and modify the admin user list, but we could add more functionality in the future.


rartiss55 (approved):

analysis-ui/src/components/Account/adminPage.jsx
```diff
@@ -0,0 +1,110 @@
+import React, {useState} from 'react';
+import {withRouter} from 'react-router-dom';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+import '../../css/adming-page.css';
```
rartiss55, line 5: Typo in filename for admin-page.css

analysis-ui/src/components/App/index.jsx
```diff
@@ -149,7 +149,12 @@ function Admin({newState, userLoginHandler}) {
         history.push("/login");
     }
 
-    return <AdminPage currentUser={newState.currentUser} updateUserHandler={userLoginHandler}/>
+    // Do not let users who aren't admins somehow go to the admin page
+    if(newState.currentUser !== null && newState.currentUser.admin === true) {
```
rartiss55, line 153: Nice catch!

rartiss55:
This looks great -- only noticed this on a second pass, if you use the admin page to modify users and leave the page/come back without doing a hard refresh, the lists don't stay updated (its probably just caching the previous getUsers query results from before the modification).

analysis-ui/src/components/Account/adminPage.jsx
brianpippin: Good catch, fixed the names.

brianpippin:
> This looks great -- only noticed this on a second pass, if you use the admin page to modify users and leave the page/come back without doing a hard refresh, the lists don't stay updated (its probably just caching the previous getUsers query results from before the modification).

It's suppose to update the user as soon as you move it back and forth, but I will check to make sure the getUsers call isn't caching, will test this some more before merging.

PR #48, opened 2021-08-16
Save sorting and grouped by options to saved query

jaudick:
Run these first to make it work. I also updated the csv naming. I had to change some functional components to class components to forward the ref because the ref would become undefined if they were different. 
```
docker build --tag node-graphql node-graphql/.
docker build --tag node-graphql-staging node-graphql/. --build-arg PORT_ARG=9111
```

### DIFF
--- a/analysis-ui/src/components/QueryBuilder/queryResults.jsx
+++ b/analysis-ui/src/components/QueryBuilder/queryResults.jsx
@@ -12,70 +12,74 @@ const create_complex_query = gql`
         createComplexQuery(queryObject: $queryObject) 
     }`;
 
-const Results = ({queryObj}) => {
+class Results extends React.Component {
 
-    return (
-        <Query query={create_complex_query} variables={{"queryObject": queryObj}}>
-        {
-            ({ loading, error, data }) => {
-                if (loading) return <div>Results are currently loading ... </div> 
-                if (error) return <div>Error</div>
-                
-                let resultsData = data[complexQueryName].results;
+    render() {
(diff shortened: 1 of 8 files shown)


analysis-ui/src/components/QueryBuilder/queryResults.jsx
```diff
@@ -12,73 +12,80 @@ const create_complex_query = gql`
         createComplexQuery(queryObject: $queryObject) 
     }`;
 
-const Results = ({queryObj}) => {
+class Results extends React.Component {
```
rartiss55, line 15: So this may have been possible to accomplish without changing this to a class based component (I would think its just passing in variables/passing the appropriate state hook to/from the grandchild). 

That being said, I tested everything and it works.

analysis-ui/src/components/QueryBuilder/queryResultsTable.jsx
```diff
@@ -92,12 +92,12 @@ class QueryResultsTable extends React.Component {
...
-            groupBy: "",
-            sortBy: "",
-            sortOrder: "asc",
+            groupBy: this.props.groupBy.value,
+            sortBy: this.props.sortBy.property,
+            sortOrder: this.props.sortBy.sortOrder,
             expandedGroups: [],
             page: 0,
-            rowsPerPage: 10
+            rowsPerPage: 10,
```
rartiss55, line 100: extra comma here

brianpippin (approved):

PR #49, opened 2021-08-19
Load Query Search Bar

jaudick:
I also changed the load query to be 50 percent of viewport height because I felt it was too small especially on larger montiors

I realized halfway through this that this ticket was supposed to be for the actual table and not load query. So I changed that old ticket to reflect this load query search bar change and made a new ticket for the original results table task here: https://nextcentury.atlassian.net/browse/MCS-856?atlOrigin=eyJpIjoiNWRmNTlkZjExY2I4NDg0OGE2ZjkzOWQ4OThhYTM4NjEiLCJwIjoiaiJ9

### DIFF
--- a/analysis-ui/src/components/QueryBuilder/loadQuery.jsx
+++ b/analysis-ui/src/components/QueryBuilder/loadQuery.jsx
@@ -17,8 +17,20 @@ const LOAD_SAVED_QUERIES = gql`
         }
     }`;
 
+function LoadQuerySearchBar({setSearch}) {
+    return (
+        <div className="load-query-search-container">
+            <span className="material-icons icon-margin-left" style={{paddingRight: "5px", paddingTop: "7px", borderBottom: "1px solid", borderBottomStyle:"inset"}}>
+                    search
+            </span>
+            <input className="load-query-search-bar" type="text" id="loadQuerySearchBar" placeholder="Search..." onChange={(e)=>setSearch(e.target.value)}/>
+        </div>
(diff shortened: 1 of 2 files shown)


rartiss55 (approved):
I like this change even though it wasn't the original intention of the ticket.

analysis-ui/src/components/QueryBuilder/loadQuery.jsx
```diff
@@ -17,15 +17,28 @@ const LOAD_SAVED_QUERIES = gql`
         }
     }`;
 
+function LoadQuerySearchBar({setSearch}) {
+    return (
+        <div className="load-query-search-container">
+            <span className="material-icons icon-margin-left" style={{paddingRight: "5px", paddingTop: "7px", borderBottom: "1px solid", borderBottomStyle:"inset"}}>
```
brianpippin, line 23: It would be nice to move this styling to a class instead of inline, I know it's a little bit of a nitpick, but having all the style in the css files, really makes the code cleaner, and easier to find where the actual styles are when needing to change them in the future.

brianpippin:
Looks great and works great.  I was wondering if you could add this to the ticket, but could you fix up the All Queries tab so there is not a horizontal scrollbar.  If you get a fix for that, and then move some of the styles into the CSS file this is ready to merge.  Awesome work.

jaudick:
@brianpippin updated

brianpippin (approved):

PR #50, opened 2021-08-20
Query Results Table Search Bar

jaudick:
I looked at the zeplin page as a reference. I thought I should add extra functionality on top of what was there. The "**All Rows**" button searches all of the rows in the table, not just the 5, 10, or 25 currently displayed in the table. I added a "**Search"** button because typing the search and updating the searches realtime with each letter input would cause the application to freeze as it looped though all the options each input. The categories limits the search to a specific category. Searches also work when the table is grouped.

### DIFF
--- a/analysis-ui/src/components/QueryBuilder/queryResultsTable.jsx
+++ b/analysis-ui/src/components/QueryBuilder/queryResultsTable.jsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import Table from "@material-ui/core/Table";
 import TableBody from "@material-ui/core/TableBody";
 import TableCell from "@material-ui/core/TableCell";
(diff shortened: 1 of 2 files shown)


brianpippin:
I think Rachel and I are going to have to look at this one next week.  I noticed there are some major peformance issues when searching all rows with like 4500 rows.  We might need to come up with some mechanism to improve this as I can see issues that would cause the UI to crash.  Let's keep this branch open, as I think it's an amazing start with things, and then we can figure out the performance issues and get it across the finish line.

jaudick:
@brianpippin Thanks! I definitely agree, performance takes a huge hit with 1000+ rows. I added a quick css styling fix with my latest commit.

brianpippin:
Closing this pull request because we closed the corresponding ticket, do not think we are going to be doing searching across the table any more.

PR #52, opened 2021-10-13
MCS-559:  Improve Home page performance - redesign API

brianpippin:


analysis-ui/src/components/Home/chartContainer.jsx
```diff
@@ -0,0 +1,159 @@
...
+            isPlausibility: this.getPlausibility(props.chartOptions[0]),
+            isWeighted: this.props.testType.toLowerCase() === 'interactive'? false : true
+        }
+
+        this.toggleChartOptions = this.toggleChartOptions.bind(this);
+        this.handleWeightedToggle = this.handleWeightedToggle.bind(this);
+    }
+
+    getMetadataLevel(valueToCheck) {
+        const optionArray = valueToCheck.value.toLowerCase().split("by");
```
ThomasSchellenbergNextCentury, line 31: Out of curiosity, what are the possible options? I'd expect "Level 1", "Level 2", "Oracle", etc. but none of those have "by" in them.

analysis-ui/src/components/Home/chartContainer.jsx
```diff
@@ -0,0 +1,159 @@
...
+        return optionArray[0];
+    }
+
+    getPlausibility(valueToCheck) {
+        if(valueToCheck.value.toLowerCase().indexOf("expected") > 0 || valueToCheck.value.toLowerCase().indexOf("plausibility") > 0) {
+            return true;
+        } else {
+            return false;
+        }
```
ThomasSchellenbergNextCentury, line 40: 1. This could just be `return <case>` rather than `if <case> return true, else return false`
2. Just confirming that this should indeed be using "plausibility" and not "plausible"

PR #56, opened 2021-11-02
MCS-511 - Plausibility Graph 

rartiss55:
Resurrecting old plausibility chart (thanks to Brian for already creating that component). Also made videos smaller to fit plausibility chart in at the top per mockups. May need to revisit sizing later depending on how the data looks. 

This was branched from the navigation changes, hence the base branch.

### DIFF
--- a/analysis-ui/src/components/Analysis/plausabilityGraph.jsx
+++ b/analysis-ui/src/components/Analysis/plausabilityGraph.jsx
@@ -6,7 +6,7 @@ import $ from 'jquery';
 const MyResponsiveLine = ({ data }) => (
     <ResponsiveLine
         data={data}
-        margin={{ top: 20, right: 175, bottom: 30, left: 80 }}
+        margin={{ top: 20, right: 100, bottom: 50, left: 80 }}
         xScale={{ type: 'linear', min: 1, max: 61}}
         yScale={{ type: 'linear', min: 0, max: 1.1}}
         axisBottom={{
@@ -14,8 +14,8 @@ const MyResponsiveLine = ({ data }) => (
             tickSize: 5,
             tickPadding: 5,
             tickRotation: 0,
-            legend: 'Frame',
-            legendOffset: 36,
+            legend: 'Step/Frame',
+            legendOffset: 35,
             legendPosition: 'middle',
             tickValues: [10, 20, 30, 40, 50, 60]
         }}
@@ -72,7 +72,7 @@ const MyResponsiveLine = ({ data }) => (
 function PlausabilityGraph ({pointsData}) {
 
     return (
-        <div style={{ height: '300px', width: '100%' }}>
+        <div style={{ height: '350px', width: '400px' }}>
             <MyResponsiveLine data={pointsData} />
         </div>
     )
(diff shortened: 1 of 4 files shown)


rartiss55:
@ThomasSchellenbergNextCentury suggested screenshots would be helpful....

Full page view
Screen Shot 2021-11-02 at 2 24 15 PM

Here's a closeup example for one of the Eval 3 agency tasks (the gaps are "EndHabituation" calls):
Screen Shot 2021-11-02 at 2 24 37 PM

analysis-ui/src/components/Analysis/plausabilityGraph.jsx
```diff
@@ -66,56 +65,34 @@ const MyResponsiveLine = ({ data, state, onClickHandler}) => (
...
-                            tempData[i].data = $.extend(true, [], this.props.pointsData[j].data);
-                            $("#scene_image_" + sceneNumber).css({opacity: 1});
-                        }
-                    }
-                }
-            }
+    const calculateXTicks = () => {
+        // int phys tasks currently always have 60 frames and
+        // therefore preset tick values, but agency scenes can vary. 
+        if(xAxisMax === INT_PHYS_MAX_FRAMES) {
```
ThomasSchellenbergNextCentury, line 82: The number of frames would be the same as the number of steps, because we're returning an image after each step, right? In that case, a constant 60 frames is no longer correct for the Eval 4 tasks. The Eval 4 intuitive physics scenes will have between 160 and 240 frames/steps, depending on the task.

ThomasSchellenbergNextCentury: If it's easy to grab the last_step from the scene's goal, then you could use that value.

rartiss55: Thanks for pointing this out -- I updated it so that it does the calculation it does in the else statement every time. 

Re: last_step: I'm using the length of the steps array in the history file instead, which should be okay (there were some edge cases where people went beyond the last_step value, but I do think that's information we want to know/visualize). I also changed the label on the graph to say "Step" instead of "Frame" because of this.

analysis-ui/src/components/Analysis/scenes.jsx
```diff
@@ -299,6 +300,48 @@ class Scenes extends React.Component {
             && this.getSceneHistoryItem(scenesByPerformer)["category"] === "interactive";
     }
 
+    getPointsData(sceneHistItem) {
+        let data = [];
+
+        sceneHistItem.steps.forEach(step => {
+            let plausibility = step["confidence"];
```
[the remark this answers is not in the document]
rartiss55: I need to review this again -- was looking at the docs again last night and I think this might be a bit off.

analysis-ui/src/components/Analysis/plausabilityGraph.jsx
ThomasSchellenbergNextCentury: Sounds great!

analysis-ui/src/components/Analysis/scenes.jsx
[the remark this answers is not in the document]
rartiss55: Okay, edited the rules above. Should be right now @ThomasSchellenbergNextCentury if you wouldn't mind taking a second look.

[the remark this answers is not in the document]
ThomasSchellenbergNextCentury: Good catch. Looks good to me.

ThomasSchellenbergNextCentury (approved):

PR #58, opened 2021-11-10
MCS-979:  Remove paired scoring view for agency and ability to view by 'expected'

brianpippin:
Only view agency by paired, the other view doesn't give a correct indication of the results with how the tasks were designed.


analysis-ui/src/components/Home/chartContainer.jsx
```diff
@@ -127,7 +127,7 @@ class ChartContainer extends React.Component {
                             defaultValue={this.state.chartOption}
                         />
                     </div>
-                    {this.props.testType.toLowerCase() !== 'interactive' &&
+                    {this.props.testType.toLowerCase() === 'intuitive physics' &&
```
brianpippin, line 130: Didn't make sense to do not equal to interactive and not equal to agents now, so just changed it to be equal to intuitive physics.

ThomasSchellenbergNextCentury (approved):

PR #63, opened 2021-11-19
Analysis query fix

rartiss55:
Found this a couple days ago when playing with dev graphql -- it looked like there were duplicates at first, but then realized I likely need an extra clause to specify which scene data to pull since we're sharing names/descriptions for things in Eval 4. 

Now that I'm looking at this again, I'm wondering if there's something with the complex query itself that should be updated instead (I was able to do something similar in dev on the Query Builder page to demonstrate, if you examine the scene info on the query I saved in dev for "MESS golf Eval 4 Results Query", but that could also be user error since I only used history collection fields and not one from "scene" as well).


ThomasSchellenbergNextCentury (approved):
So this helps to filter the query results by the currently selected eval/"Collection" (e.g. Eval 3 vs Eval 4)?

brianpippin (approved):

PR #65, opened 2021-11-22
MCS-1009 Fix scene history query on analysis UI page

rartiss55:
Abandoning using createComplexQuery on the scene analysis page and switching to a more accurate query for scene history (using eval, category_type, and test_num).

### DIFF
--- a/analysis-ui/src/components/Analysis/scoreTable.jsx
+++ b/analysis-ui/src/components/Analysis/scoreTable.jsx
@@ -47,13 +47,13 @@ function ScoreTable({columns, currentPerformerScenes, currentSceneNum,
                     <TableRow>
                     {columns.map((col, colKey) => (
                         <TableCell key={"performer_score_header_cell_" + colKey}>
-                            {sortable &&
+                            {sortable && col.dataType !== 'scene' &&
                                 <TableSortLabel active={sortOption.sortBy === col.dataKey} direction={sortOption.sortOrder} 
                                     onClick={() => handleRequestSort(col.dataKey)}>
                                     {col.title}
                                 </TableSortLabel>
                             }
-                            {!sortable &&
+                            {((!sortable) || col.dataType === 'scene') &&
                                 col.title
                             }
                         </TableCell>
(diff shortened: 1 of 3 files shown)


node-graphql/server.schema.js
```diff
@@ -62,6 +62,8 @@ const mcsTypeDefs = gql`
     metadata: String
     filename: String
     fileTimestamp: String
+    fullFilename: String
+    corner_visit_order: [JSON]
```
rartiss55, line 66: We were bypassing updating this before by using the complex query (hopefully my type is correct for corner_visit_order, I don't have good enough local data to test but this was my assumption based on what's in ingest).

analysis-ui/src/components/Analysis/scoreTable.jsx
```diff
@@ -47,13 +47,13 @@ function ScoreTable({columns, currentPerformerScenes, currentSceneNum,
...
                         <TableCell key={"performer_score_header_cell_" + colKey}>
-                            {sortable &&
+                            {sortable && col.dataType !== 'scene' &&
                                 <TableSortLabel active={sortOption.sortBy === col.dataKey} direction={sortOption.sortOrder} 
                                     onClick={() => handleRequestSort(col.dataKey)}>
                                     {col.title}
                                 </TableSortLabel>
                             }
-                            {!sortable &&
+                            {((!sortable) || col.dataType === 'scene') &&
```
rartiss55, line 56: Since the data for slices comes from a different source than what's actually in the "history" result now, turning off sorting for those columns.

node-graphql/server.schema.js
brianpippin: Something is up with the corner visit order not appearing in the UI, I am going to attempt a fix and will push when I get it to appear.

brianpippin: Okay they appear now, check out the fix in scenes.jsx (old line 79, current line 76 i believe)

brianpippin (approved):

node-graphql/server.schema.js
rartiss55: Makes sense to me 👍

kdrumm (approved):
Code looks good to me

PR #72, opened 2021-11-29
MCS-991: Make sure Eval is always in reverse order, so latest eval is at top

brianpippin:


ThomasSchellenbergNextCentury (approved):
Yup, that's a sort. :smiley:

kdrumm (approved):

PR #73, opened 2021-11-30
MCS-1028: Add ability to do an equals with an input field

brianpippin:
Add to query builder an Input field for equals - make default to improve load speed

### DIFF
--- a/analysis-ui/src/components/QueryBuilder/queryLine.jsx
+++ b/analysis-ui/src/components/QueryBuilder/queryLine.jsx
@@ -7,7 +7,8 @@ import { Query } from 'react-apollo';
 import Select from 'react-select';
 
 const functionOptions = [
-    {value: "equals", label: "equals"},
+    {value: "equalsInput", label: "equals (input)"},
+    {value: "equalsDropdown", label: "equals (dropdown)"},
     {value: "contains", label: "contains"},
     {value: "does_not_contain", label: "does not contain"},
     {value: "between", label: "between"},
@@ -89,7 +90,7 @@ class QueryLineItem extends React.Component {
             fieldNameLabel: "",
             fieldValue1: "",
             fieldValue2: "",
-            functionOperator: "equals",
+            functionOperator: "equalsInput",
             collectionDropdownToggle: 0
         };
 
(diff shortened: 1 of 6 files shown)


node-graphql/server.statsFunctions.js
```diff
@@ -50,14 +50,8 @@ function getChartOptions(evalType, metadata, hasNovelty) {
...
 function updateOverall(statsObj, scoreStatsObj, performer, weight) {
     if(statsObj.hasOwnProperty(performer)) {
-        if(weight === 0) {
-            console.log(statsObj[performer] + (scoreStatsObj["count"] * weight));
-        }
         statsObj[performer] = statsObj[performer] + (scoreStatsObj["count"] * weight);
     } else {
-        if(weight === 0) {
-            console.log(scoreStatsObj["count"] * weight);
-        }
```
brianpippin, line 60: This is just removing unnecessary console logs that must have been for debugging but are no longer necessary and are clogging up the docker logs for node-graphql container.

ThomasSchellenbergNextCentury (approved):

deanwetherby (approved):

analysis-ui/src/components/QueryBuilder/queryLine.jsx
```diff
@@ -7,7 +7,8 @@ import { Query } from 'react-apollo';
 import Select from 'react-select';
 
 const functionOptions = [
-    {value: "equals", label: "equals"},
+    {value: "equalsInput", label: "equals (input)"},
+    {value: "equalsDropdown", label: "equals (dropdown)"},
```
rartiss55, line 11: Does this break backwards compatibility with a previously saved query/will those not reload properly? Wondering if something needs to be added to treat previously saved as "equals" queries as "equalsDropdown" or something like that.

brianpippin: Excellent question, I might just need to leave the equals functionality server side, so that it works for existing queries but you can't add it anymore.  I will look into that next.

brianpippin:
@rartiss55 - I checked some old queries and this fix appears to be backwards compatible, let me know if you see anything else.

rartiss55 (approved):
@brianpippin LGTM!

PR #77, opened 2021-12-15
Add link to log file to UI (MCS-887).

rartiss55:
Linking to the most recent log file (matching the history file timestamp) in the UI.

### DIFF
--- a/analysis-ui/src/components/Analysis/evalConstants.jsx
+++ b/analysis-ui/src/components/Analysis/evalConstants.jsx
@@ -1,10 +1,16 @@
 import {RESOURCES_URL} from '../../services/config';
 
+const baseEvalConstant = {
+    movieExtension: ".mp4",
+    sceneExtension: "_debug.json",
+    logExtension: "_log.txt" // note that there are only logs for Eval 4+
+}
+
 const Eval2 = {
+    ...baseEvalConstant,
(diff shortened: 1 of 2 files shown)


rartiss55:
Forgot to add -- we are saving all the log files if there are multiple ones/if we ever do need them (since they are timestamped correctly).

deanwetherby (approved):

analysis-ui/src/components/Analysis/evalConstants.jsx
```diff
@@ -51,6 +51,7 @@ const Eval4 = {
     movieExtension: ".mp4",
     sceneBucket: RESOURCES_URL + "/eval-scenes-4/",
     sceneExtension: "_debug.json",
+    logExtension: "_log.txt"
```
brianpippin, line 54: I think this might be a good time to use refactoring, and perhaps make a base object with things are consistent across every evaluation, and then use the ... notation to combine the base object with the eval specific properties.  Thoughts?

rartiss55: Done...also removed some of the extra movie extension properties since they're only needed for eval 2 data. LMK if I missed anything.

brianpippin (approved):
It looks so much cleaner now in the file, Thanks for doing that so quickly Rachel.  Looks great!

PR #80, opened 2021-12-21
MCS-1061: Add do not answer to home page calculations, with toggle

brianpippin:


deanwetherby (approved):
It has to be tough using camelCase here and snake_case in our Python code. camelCase looks weird to me now.

brianpippin:
> It has to be tough using camelCase here and snake_case in our Python code. camelCase looks weird to me now.

Yeah it can be tricky to keep it all straight, but also css classes uses "-" between so it's like modified snake-case.

kdrumm (approved):

PR #85, opened 2022-02-03
MCS-835: Update node server dependencies

brianpippin:
All of the warnings have gone away except for a few.  Despite using the latest version of the aws-sdk dependency, that dependency is causing some warnings and we can not make those go away until aws-sdk updates their own dependencies. 

Also removed dependencies that were not being used.  This code is based off of MCS-841 because that is the state of my database but these changes could be applied to development as well and should work.


ThomasSchellenbergNextCentury (approved):
I don't currently have a good way of testing this, so I don't know if my approval is any use, but here you go.

PR #86, opened 2022-02-07
MCS-636:  Add FPS to UI

brianpippin:
This ticket adds FPS and delta milliseconds (real time between actions) to the step table.  This also makes that table visible for interactive however it excludes columns that are only pertinent for intuitive physics from the interactive view.

NOTE:  This ticket is based on MCS-841 so it will merge sometime after the PI Meeting.


deanwetherby (approved):

ThomasSchellenbergNextCentury (approved):

rartiss55 (approved):
I'm okay leaving it as is but if its in the data now, perhaps InteractiveScenePlayer should be updated to use that value (its hardcoded as 20 FPS there). Also if FPS becomes more variable (different per step) we may need to revisit that implementation in the future.

PR #89, opened 2022-05-31
Move csv check to node-graphql using url-exist dependency (MCS-935).

rartiss55:
This was to prevent polluting the console with 404-related errors if a csv file hasn't yet been generated.

### DIFF
--- a/analysis-ui/src/components/EvaluationStatus/evalStatusTable.jsx
+++ b/analysis-ui/src/components/EvaluationStatus/evalStatusTable.jsx
@@ -1,5 +1,5 @@
 import React, {useState, useEffect} from 'react';
-import { Query } from 'react-apollo';
+import { Query, useQuery } from 'react-apollo';
 import gql from 'graphql-tag';
 import Select from 'react-select';
 import EvalStatusConfigureModal from './evalStatusConfigure';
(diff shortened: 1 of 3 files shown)


analysis-ui/src/components/EvaluationStatus/evalStatusTable.jsx
```diff
@@ -66,11 +72,15 @@ function CreateCSV() {
...
 
 function CSVDownloadLink({url, linkText}) {
+    const {data, refetch} = useQuery(get_link_status, {variables: {url}, fetchPolicy: 'no-cache'});
     const [linkActive, updateLinkActive] = useState();
+
     useEffect(() => {
         const getUrl = async () => {
-            const linkCheck = await fetch(url);
-            updateLinkActive(linkCheck.ok);
+            refetch({variables: {url}});
```
rartiss55, line 80: Something I thought of while working on this ticket: once the csv file exists, do we need to keep polling to check the URL? I know we regenerate the csv from time to time but the URL itself doesn't change, right?

brianpippin: The main problem isn't that we are attempting to poll for the url constantly, it is that the page is "refreshing" with new test results periodically, so the component is being rerendered and that is what is causing the constant csv check as well.  Perhaps we need a way to turn off polling or mark the eval complete?

brianpippin (approved):
Code looks good.  Could be a new ticket to turn on and off polling, especially if Eval is complete.

ThomasSchellenbergNextCentury (approved):
Haven't tested it myself, but the code changes look fine.

jaudick (approved):

analysis-ui/src/components/EvaluationStatus/evalStatusTable.jsx
rartiss55: Added a ticket here: https://nextcentury.atlassian.net/browse/MCS-1370

PR #90, opened 2022-06-02
Remove rating/score columns for interactive scenes (MCS-621).

rartiss55:
Analysis UI page update: removing rating and score columns from score table for interactive scenes since they don't apply there.

### DIFF
--- a/analysis-ui/src/components/Analysis/scenes.jsx
+++ b/analysis-ui/src/components/Analysis/scenes.jsx
@@ -47,6 +47,7 @@ const mcs_history = gql`
             scene_num
             scene_goal_id
             score
+            slices
             steps
             flags
             metadata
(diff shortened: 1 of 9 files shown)


analysis-ui/src/components/Analysis/scenes.jsx
```diff
@@ -74,7 +74,14 @@ const scoreTableCols = [
     { dataKey: 'score.confidence', title: 'Score/Confidence', dataType: 'history'}
 ]
 
-const scoreTableColsWithCorners = scoreTableCols.concat([{ dataKey: 'corner_visit_order', title: 'Corner Visit Order', dataType: 'history'}])
+const scoreTableColsInteractive = [
```
rartiss55, line 77: I think hardcoding the values in the score table component instead might be preferable, but wanted to hold off on refactoring that until slices were available on the history records so I can clean some of that logic up too (we can get rid of the dataType key here in that case).

brianpippin: I just merged the code for the script that addes slices, if you want to run that locally so you can finish up the part you mentioned.

brianpippin (approved):
Code looks good.  Will rereview if you add the new slice stuff or create a new ticket for that if you want.

ThomasSchellenbergNextCentury (approved):

jaudick (approved):

rartiss55:
> Code looks good. Will rereview if you add the new slice stuff or create a new ticket for that if you want.

Thanks @brianpippin -- added the slice stuff in and nixed some of the refactoring I had in mind since I forgot about the eval 2 case. Also added minor fixes to get the eval 2 analysis page working again.

analysis-ui/src/components/TestOverview/evalNavItem.jsx
```diff
@@ -61,7 +61,8 @@ class EvalNavItem extends React.Component {
 
                     // Remove Evaluation 2 Results because this page is designed to show hypercube results
                     //    and we did not have a hypercube design for Evaluation 2
-                    evalOptions = evalOptions.filter(element => element.label !== "Evaluation 2 Results");
+
+                    evalOptions = evalOptions.filter(element => element.label !== "eval_2_results");
```
brianpippin, line 65: I have to pull this out and test, but I think the label will still be "Evaluation 2 Results" and the value will be "eval_2_results" we should test this one.

rartiss55: Ah, you're right! Fixed now.

rartiss55: Ah, you're right! Fixed now.

brianpippin (approved):

PR #92, opened 2022-06-02
MCS-1198: Add view to test overview to be able to view by slice and slice levels

brianpippin:
Modified Test Overview to have tabs for cube id/slice/scorecard - only show tab and table if applicable to that test/category type.  Made it so on the slice view, you can view by slice levels.

### DIFF
--- a/analysis-ui/src/components/TestOverview/hypercubeResultsTable.jsx
+++ b/analysis-ui/src/components/TestOverview/hypercubeResultsTable.jsx
@@ -7,13 +7,14 @@ import TableBody from "@material-ui/core/TableBody";
 import TableCell from "@material-ui/core/TableCell";
 import TableHead from "@material-ui/core/TableHead";
 import TableRow from "@material-ui/core/TableRow";
+import { default as ReactSelect } from "react-select";
+import { components } from "react-select";
 
 const hyperCubeDataQueryName = "getTestOverviewData";
 const getHyperCubeData = gql`
 
-query getTestOverviewData($eval: String!, $categoryType: String!, $performer: String!, $metadata: String!, $useDidNotAnswer: Boolean!, $weightedPassing: Boolean!, $statType: String!, $sliceLevel: Int!) {
-    getTestOverviewData(eval: $eval, categoryType: $categoryType, performer: $performer, metadata: $metadata, useDidNotAnswer: $useDidNotAnswer, weightedPassing: $weightedPassing, statType: $statType, sliceLevel: $sliceLevel) 
-
+query getTestOverviewData($eval: String!, $categoryType: String!, $performer: String!, $metadata: String!, $useDidNotAnswer: Boolean!, $weightedPassing: Boolean!, $statType: String!, $sliceLevel: Int!, $sliceType: String!, $sliceKeywords: JSON!) {
+    getTestOverviewData(eval: $eval, categoryType: $categoryType, performer: $performer, metadata: $metadata, useDidNotAnswer: $useDidNotAnswer, weightedPassing: $weightedPassing, statType: $statType, sliceLevel: $sliceLevel, sliceType: $sliceType, sliceKeywords: $sliceKeywords) 
 }`;
 
 const overViewTableFieldsStatic = [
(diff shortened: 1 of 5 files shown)


analysis-ui/src/components/TestOverview/hypercubeResultsTable.jsx
```diff
@@ -76,31 +50,48 @@ class HyperCubeResultsTable extends React.Component {
...
                             </div>
+                            {this.props.hyperCubePivotValue === "slice" &&
+                                <div className="overview-results-slice-chooser">
+                                    <span className="slice-header">Slice level: </span>
+                                    <div className="btn-group me-2" role="group">
+                                        <button type="button" className={this.state.sliceLevel === 1 ? activeButtonClasses : notActiveButtonClasses} onClick={()=>this.updateSliceLevel(1)}>1</button>
+                                        <button type="button" className={this.state.sliceLevel === 2 ? activeButtonClasses : notActiveButtonClasses} onClick={()=>this.updateSliceLevel(2)}>2</button>
+                                        <button type="button" className={this.state.sliceLevel === 3 ? activeButtonClasses : notActiveButtonClasses} onClick={()=>this.updateSliceLevel(3)}>3</button>
+                                        <button type="button" className={this.state.sliceLevel === 4 ? activeButtonClasses : notActiveButtonClasses} onClick={()=>this.updateSliceLevel(4)}>4</button>
```
ThomasSchellenbergNextCentury, line 91: Do the number of "slice levels" correspond to the dimensionality of the particular hypercube? So a 2D hypercube would have 2 slice levels, and a 4D hypercube would have 4?

node-graphql/server.statsFunctions.js
```diff
@@ -382,40 +382,65 @@ function updateStatObj(hyperCubeObj, statObj) {
...
+            if(i < sliceArrayLength -1) {
+                statStr += ", ";
+            }
+        }
+
+        return statStr;
+    } else {
+        return statName;
+    }
+}
```
ThomasSchellenbergNextCentury, line 407: Maybe this is just due to my unfamiliarity with the UI, but an example of what this function returns (a `statStr` and `statName`) might be helpful here?

analysis-ui/src/components/TestOverview/hypercubeResultsTable.jsx
brianpippin: It contains to like the setup and the dimensions.  I've attached screenshots of obj perm
Screen Shot 2022-06-03 at 12 23 51 PM
Screen Shot 2022-06-03 at 12 24 01 PM
Screen Shot 2022-06-03 at 12 24 07 PM
Screen Shot 2022-06-03 at 12 24 13 PM

Does that help show what is going on?  If the team wants to rename the UI label (psychs as well we can do that, plan to show Koleen, who will probably the final judge if this is what she wanted.

node-graphql/server.statsFunctions.js
brianpippin: This function will either return the hypercubeId (so "A1" or "A2", or the slice level that shows so the first column in the screen shot I sent you on the other comment.  I will look to add some more comments to the code to make that more clear.

analysis-ui/src/components/TestOverview/hypercubeResultsTable.jsx
ThomasSchellenbergNextCentury: Ah, OK, that does help, thanks! And I'm not sure that I can think of a better label.

Can I not configure specific combinations right now? For example, "setup move across whole scene, object one novelty untrained shape"? That's only two slices, but it looks like, to get novelty in this example, I would always have to go to slice level 4?

I don't want to trivialize the usefulness of the current page -- it still looks super helpful -- just thinking out loud.

brianpippin: At the moment you can not configure things at that level, I think that is a question I have for KDL with how they plan to use the page.  Trying to configure it at that setup level would require a totally different design, which is what I was hoping to get from them.  Which is why this won't get merged until KDL signs off on it.  When I was looking at the graphs that they build, it seems like they build up from cube level, but it might be this doesn't give them enough flexibility.  To be determined.

ThomasSchellenbergNextCentury: Sounds good. Thanks for the explanation!

brianpippin:
After meeting with psychs today and going over things, this is ready for a final review.

PR #93, opened 2022-06-07
Load Query Redesign

jaudick:
First time touching React since August so I was a little rusty but I think this turned out good. I added a bonus of rows being highlighted when you hover over them so you know that you are going to select it and that you can click the entire row, not just the checkbox to make a selection. David wanted alternating rows to be  slightly different colors so that's the case for both unselected (white/light-grey) and selected rows (light-blue/dark-blue).

Selecting one query will replace the current query tab. Selecting more than one query will append multiple queries to the end of the tab list.

Here is a screenshot of the original design idea from zeplin that I tried to match.
Screen Shot 2022-05-24 at 4 54 31 PM (1)

### DIFF
--- a/analysis-ui/src/components/QueryBuilder/loadQuery.jsx
+++ b/analysis-ui/src/components/QueryBuilder/loadQuery.jsx
@@ -35,9 +35,9 @@ function LoadQueryPage({show, onHide, currentUser, loadQueryHandler}) {
     const [search, setSearch] = useState("");
     const [selectedQueries, setSelectedQueries] = useState([]);
 
-    const loadQuery = (query) => {
-        loadQueryHandler(query);
-        onHide();
+    const loadQuery = () => {
+        let queries = selectedQueries
+        loadQueryHandler(queries);
     }
 
     const closeModal = () => {
@@ -103,7 +103,7 @@ function LoadQueryPage({show, onHide, currentUser, loadQueryHandler}) {
                                 <LoadQuerySearchBar setSearch={setSearch}/>
                                 <span>
                                     <p>({selectedQueries.length}) Selected</p>
-                                    <button type="button" onClick={() => loadQuery("")}>Load Selected</button>
+                                    <button type="button" onClick={() => loadQuery()}>Load Selected</button>
                                 </span>
                             </div>
                             <table className="load-query-table">
--- a/analysis-ui/src/components/QueryBuilder/loadQuery.jsx
+++ b/analysis-ui/src/components/QueryBuilder/loadQuery.jsx
@@ -64,18 +64,8 @@ function LoadQueryPage({show, onHide, currentUser, loadQueryHandler}) {
     }
 
     useEffect(() => {
-        let checkbox = document.getElementById('header_checkbox');
-        if (checkbox !== null)
-            checkbox.checked = selectedQueries.length > 0;
-    }, [selectedQueries])
-
-    useEffect(() => {
         updateSelectedBasedOnSearch();
-    }, [search])
-
-    useEffect(() => {
-        updateSelectedBasedOnSearch();
-    }, [activeTab])
+    }, [search, activeTab])
 
     const setSelected = (key, button) => {
         const checkbox = document.getElementById('load_query_checkbox_' + key);
@@ -153,7 +143,7 @@ function LoadQueryPage({show, onHide, currentUser, loadQueryHandler}) {
                                         <th style={{columnWidth: '1vw'}}>
                                             <a href='#' data-toggle="tooltip" title="Checking this box will select all of the queries matching the search parameters. 
                                             If there are no search parameters, all queries will be selected. Unchecking this box will clear all selections.">
-                                                <input type="checkbox" id="header_checkbox" onClick={(e) => selectOrClearAll(e)}/>
+                                                <input type="checkbox" id="header_checkbox" checked={selectedQueries.length > 0} onClick={(e) => selectOrClearAll(e)}/>
                                             </a>
                                         </th>
                                         <th style={{columnWidth: '50vw'}}>Title</th>
--- a/analysis-ui/src/components/QueryBuilder/loadQuery.jsx
+++ b/analysis-ui/src/components/QueryBuilder/loadQuery.jsx
@@ -1,5 +1,4 @@
-import React, {useState, useEffect} from 'react';
-import Modal from 'react-bootstrap/Modal';
+import React, {useState} from 'react';
 import gql from 'graphql-tag';
 import { Query } from 'react-apollo';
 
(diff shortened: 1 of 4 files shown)


analysis-ui/src/components/QueryBuilder/loadQuery.jsx
```diff
@@ -30,114 +30,178 @@ function LoadQuerySearchBar({setSearch}) {
...
+                                    </tr>
+                                </thead>
+                                <tbody>
+                                    {
+                                        queries.map((query, key) =>
+                                        (search === "" ||
+                                        (query.name.toLowerCase().includes(search.toLowerCase()) || 
+                                        query.description.toLowerCase().includes(search.toLowerCase()) || 
+                                        query.user.username.toLowerCase().includes(search.toLowerCase()))) &&
+                                        <tr key={'load_query_row_' + key} id={'load_query_row_' + key} onClick={() => setSelected(key, false)} query={JSON.stringify(query)}>
```
jaudick, line 176: `query={JSON.stringify(query)}`
Is defining my own custom html attribute ok here? And it seemed that making it query={{query}} didnt work as it always made it [object Object] instead of the actual query object and I could never extract the properties.

analysis-ui/src/components/QueryBuilder/loadQuery.jsx
```diff
@@ -30,114 +30,178 @@ function LoadQuerySearchBar({setSearch}) {
...
+                                    <a href='#selected' data-toggle="tooltip" 
+                                    title='Selecting one query will replace the current query tab. Selecting more than one query will append mutiple queries to the end of the tab list.'>({selectedQueries.length}) Selected</a>
+                                    <button type="button" onClick={() => loadQuery()}>Load Selected</button>
+                                </span>
+                            </div>
+                            <table className="load-query-table">
+                                <thead>
+                                    <tr>
+                                        <th style={{columnWidth: '1vw'}}>
+                                            <a href='#' data-toggle="tooltip" title="Checking this box will select all of the queries matching the search parameters.
```
jaudick, line 156: This is something I added. Maybe there should be better way to know this other than a tool tip?

analysis-ui/src/components/QueryBuilder/loadQuery.jsx
```diff
@@ -30,114 +30,178 @@ function LoadQuerySearchBar({setSearch}) {
...
+    const loadQuery = () => {
+        let queries = selectedQueries;
+        loadQueryHandler(queries);
     }
 
-    const closeModal = () => {
-        setSearch("");
-        onHide();
+    const manageSelectedQueries = (query, key, remove) => {
+        let length = selectedQueries.length
```
brianpippin, line 44: Missing semicolon

brianpippin: Also lines 46 and 50

brianpippin: 53 and 54 as well.  Also I don't think you want to use const for checkbox.  I know you are changing a property on checkbox, but I think it's better convention to use "let" if you are going to be changing anything about the object.

analysis-ui/src/components/QueryBuilder/loadQuery.jsx
```diff
@@ -30,114 +30,178 @@ function LoadQuerySearchBar({setSearch}) {
...
+        if (remove) {
+            length -= 1
+            setSelectedQueries(selectedQueries.filter(item => key != item.key));
+        }
+        else {
+            length += 1
```
brianpippin, line 50: I am not sure you need this length variable at all.  I feel like you could remove lines 44, 46, and 50 and change line 54 to checkbox.checked = selectedQueries.length > 0; and not really lose anything.

analysis-ui/src/components/QueryBuilder/queryPage.jsx
```diff
@@ -52,16 +53,54 @@ class QueryPage extends React.Component {
...
+                name: "Query " + (this.state.totalTab + (i - this.state.totalTab)),
+                tabQueryObj: [],
+                groupBy: {value: "", label: ""},
+                sortBy: {property: "", sortOrder: "desc"},
+                mongoId: ""
+            })
+        }
+        let newArray = this.state.queryTabs.concat(additionalTabs);
+
+        await new Promise(resolve => {
```
brianpippin, line 71: Something seems off with all the promises and awaits.  That isn't really the best way to use React state objects usually.  Can you perhaps speak to what caused you to add them?  I also am not sure why these two right here couldn't be combined together as well.

analysis-ui/src/components/QueryBuilder/loadQuery.jsx
ThomasSchellenbergNextCentury: Is there an auto-linter for the UI code that could address simple things like missing semicolons? If not, would it be worth adding one during the Post-Eval Cleanup phase?

analysis-ui/src/components/QueryBuilder/loadQuery.jsx
```diff
@@ -30,114 +30,178 @@ function LoadQuerySearchBar({setSearch}) {
...
-                                </div>   
-                            </div>
-                        )
+    useEffect(() => {
+        updateSelectedBasedOnSearch();
+    }, [search])
+
+    useEffect(() => {
+        updateSelectedBasedOnSearch();
+    }, [activeTab])
```
ThomasSchellenbergNextCentury, line 80: Newbie React question: Could these be combined?

analysis-ui/src/components/QueryBuilder/queryPage.jsx
jaudick: The use of two promises here is unnecessary and can be combined. I will change that. In fact, all of these promises here are completely unnecessary after some more testing so its good you pointed this out. 
The only thing needed is to make addMultipleTabs and loadQueryHandler both async and to `await this.addMultipleTabs(queryObjects)` and that seems to wait for the state to update as if it was a promise. Heres an explanation of the why the await is needed:

1. loadQueryHandler is called
2. addMultipleTabs is called and all of those tabs are added at once so those tabs can be loaded over with queries
3. loadQueries is called. Those queries are loaded into the new tabs. If await is not used on addMultipleTabs the state will not be updated in time for loadQueries to load those queries over those new tabs.

I pushed a new commit to remove unnecessary promises and just use await of this.addMultipleTabs while keeping that method async and loadQueryHandler async

analysis-ui/src/css/app.css
```diff
@@ -803,12 +819,90 @@ ol, ul {
...
+    border-left: 0px;
+    border-radius: 5px 5px 0px 0px;
+}
+
+.query-tab-controls button.selected {
+    background: white;
+    border-bottom: 0px;
+    border-top: 2px solid #dee2e6;
+    border-right: 2px solid #dee2e6;
+    border-left: 2px solid#dee2e6;
```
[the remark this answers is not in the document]
jaudick: Yes I will fix that

analysis-ui/src/components/QueryBuilder/loadQuery.jsx
jaudick: Thats a way better way to do it i will make that change

analysis-ui/src/components/QueryBuilder/loadQuery.jsx
jaudick: Unless Im missing something, it looks like they can after testing it! I didn't know that worked

jaudick:
@brianpippin This makes sense and I agree. I will look through the code again and see what I can change to make it more 'react'. I will let you know when I make those changes.

jaudick:
@brianpippin The code should be very react now with the same functionality. The written code itself looks a lot cleaner and more concise now because I got rid of a ton of lines. Looking at the before and after for these changes made me realize how much simpler things could be from this new react minded viewpoint.

analysis-ui/src/components/QueryBuilder/loadQuery.jsx
```diff
@@ -30,114 +29,130 @@ function LoadQuerySearchBar({setSearch}) {
...
+                            <table className="load-query-table">
+                                <thead>
+                                    <tr>
+                                        <th style={{columnWidth: '1vw'}}>
+                                            <a href='#' data-toggle="tooltip" title="Checking this box will select all of the queries matching the search parameters. 
+                                            If there are no search parameters, all queries will be selected. Unchecking this box will clear all selections.">
+                                                <input type="checkbox" id="header_checkbox" checked={selectedQueries.length > 0} onClick={(e) => selectOrClearAll(e)}/>
+                                            </a>
+                                        </th>
+                                        <th style={{columnWidth: '50vw'}}>Title</th>
```
brianpippin, line 111: Would love to see all styles as classes in a css file at some point.  If we ever clean up the css with variables.

brianpippin:
The code changes to a more react version makes the code way more readable.  Nicely done.  Just my one comment about the styles and classes and I think this will be ready to go.

jaudick:
@brianpippin The styles now use classes

brianpippin (approved):

rartiss55 (approved):

PR #94, opened 2022-06-08
Cube results table scrolls with videos

jaudick:
The table now scrolls with a fixed header. The position of the highlighted current scene will always be in view of the table window. This is helpful especially for the play all feature as the table scrolls with the videos as you play through each scene moving down the table.

### DIFF
--- a/analysis-ui/src/components/Analysis/scenes.jsx
+++ b/analysis-ui/src/components/Analysis/scenes.jsx
@@ -170,22 +170,14 @@ class Scenes extends React.Component {
                     search: searchString + sceneToUpdate
                 });
             }
-            
-            if (downOneOrUpOneScene) {
-                let sceneRow = document.getElementById("score_table_row_scene_" + sceneNum);
-                if (sceneRow !== null) {
-                    let currentScrollPosition = document.documentElement.scrollTop;
-                    let rowPosition = sceneRow.style.position;
-                    let top = sceneRow.style.top;
-                    sceneRow.style.position = 'relative';
-                    sceneRow.style.top = '-30px';
-                    sceneRow.scrollIntoView({block: 'start'});
-                    sceneRow.style.top = top;
-                    sceneRow.style.position = rowPosition;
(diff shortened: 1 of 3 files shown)


ThomasSchellenbergNextCentury (approved):
I haven't tested it myself, but the code looks fine to me.

brianpippin (changes requested):
This only brings the scene into view from the bottom part of the scroll, I think we want the highlighted one when playing all to be at the top of the scroll panel, not at the bottom.  It can be too hard to find sometimes at the bottom.  I also think it might do the table some justice to shrink the size of the fixed height as well.

analysis-ui/src/components/Analysis/scenes.jsx
```diff
@@ -170,8 +170,23 @@ class Scenes extends React.Component {
...
                 });
             }
+            
+            if (downOneOrUpOneScene) {
+                let sceneRow = document.getElementById("score_table_row_scene_" + sceneNum);
+                if (sceneRow !== null) {
+                    let currentScrollPosition = document.documentElement.scrollTop;
+                    let rowPosition = sceneRow.style.position;
+                    let top = sceneRow.style.top;
+                    sceneRow.style.position = 'relative';
```
brianpippin, line 180: Is there anyway we can clean up this function by using classes and react rerenders?

brianpippin:
Love the functionality and how it is working now, that one function just seems like it could be a little more react like.

analysis-ui/src/components/Analysis/scenes.jsx
jaudick: I pushed a commit that changes these to classes. All of my attempts at putting it in a react rerender made me lose the functionality of being able to click a row and not have the table scroll on the click. I feel like we want to keep it so clicking **does not** scroll that row to the top, but clicking the up or down one scene buttons or play all **does** scroll so I think this class change is the most optimized it can be.

PR #100, opened 2022-07-12
Query Page Delete and Update

jaudick:
You can delete multiple queries from load query if they are tied to your account. Once a query tab is loaded, if that query is tied to your account you can update or delete it alongside clearing it and saving a new one. 

Updating gives you the option to only change the name and description but not the contents if you want. This is so you can add or remove properties from your query and change the groupby and sortby without overwriting those when you update. I found this is more useful for changing groupby and sortby. 

Active query tabs that are tied to the same mongo id as the query  being updated or deleted will be modified accordingly. So if I had 10 tabs loaded and open for the query "Jacob Test 1" and deleted that query from the load query delete modal, all of those tabs will be closed. This will also happen if I delete it from the query tab's single delete modal too. Updating one of those "Jacob Test 1" tabs will also update all of the other open tabs with the same mongoId.

### DIFF
--- a/analysis-ui/src/components/QueryBuilder/loadQuery.jsx
+++ b/analysis-ui/src/components/QueryBuilder/loadQuery.jsx
@@ -1,6 +1,7 @@
 import React, {useState} from 'react';
 import gql from 'graphql-tag';
 import { Query } from 'react-apollo';
+import DeleteQuery from './deleteQuery';
 
 const getSavedQueriesName = "getSavedQueries";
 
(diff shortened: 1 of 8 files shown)


analysis-ui/src/components/QueryBuilder/loadQuery.jsx
```diff
@@ -94,8 +95,10 @@ function LoadQueryPage({show, onHide, currentUser, loadQueryHandler}) {
                             <div className="load-query-search-load-line">
                                 <LoadQuerySearchBar setSearch={setSearch}/>
                                 <span>
-                                    <a href='#selected' data-toggle="tooltip" 
+                                    <a href='#selected' data-toggle="tooltip" style={{marginRight: '-5px'}}
```
brianpippin, line 98: Super nitpick but this should be in a class, you could even just added it as .load-query-search-load-line a {} for all links under that class name or you could create a new class name.

jaudick:
@brianpippin requested changes are implemented now

brianpippin (approved):

PR #101, opened 2022-07-13
Dev to master

rartiss55:
Realized with my scorecard changes the scorecard view on test overview is broken for eval 4. Figured we should probably do this anyway :)


brianpippin (approved):
Good with this, we can always do another one Friday or early next week.

PR #102, opened 2022-07-15
Fixed eval status table refresh rate and adding empty params does not crash the application

jaudick:


analysis-ui/src/components/EvaluationStatus/evalStatusTable.jsx
```diff
@@ -11,6 +11,7 @@ const CSV_URL_SCENE_SUFFIX = "_Scenes.csv";
 const CSV_URL_RESULTS_SUFFIX = "_Results.csv";
 const EvaluationStatusQuery = "getEvaluationStatus";
 const getLinkStatusQueryName = "getLinkStatus";
+const completedEvals = ["2", "3", "3.5", "3.75", "4"];
```
brianpippin, line 14: I have only had a chance to glance at this, but we should somehow make this a collection in the database and have a way to mark the eval complete, that way we don't need to come in and modify this file after each eval, there is just a dynamic way to add an eval to the database so we don't have to change it again in the future.  Let me know if you have any questions.

jaudick: I think adding a collection is straightforward enough. How did you picture dynamically adding if an eval is complete? Would checking if every eval status configuration that all of the scenes for that test type with the selected performers and metadata level have run be good?

image

There may be issues with that approach though because this not all scenes run even for completed evals as scene here for eval 4. Do you have a better idea?

brianpippin: I would imagine admin users could have a checkbox that would submit if the evaluation is complete, and perhaps unchecking it, removes it from the list as well.  Something up where the generate csv file links are.

jaudick:
@brianpippin I added the new button to mark completed evals if you are an admin. I forgot to mention I also fixed it so if you add in your eval configuration no test type, or metadata level, or performer, the application would crash so line 60 of evalStatusConfigure is addressing that.

node-graphql/server.schema.js
```diff
@@ -635,6 +641,23 @@ const mcsResolvers = {
             });
             return {message: "CSV creation launched."};
         },
+        updateCompletedEvals: async(obj, args, context, infow) => {
```
brianpippin, line 644: Since you return the whole array of completed evals on get, I am unsure why you don't just send the whole array back when update the completed evals, and have the UI manage the array list in the state object?

jaudick: To clarify, are you saying that this line 40 of evalStatusTable:
```
    const updateCompletedEvals = async () => {
        await updateCompletedEvalsCall({ variables: {
            evalNum: currentEval.toString()
        }});
        refetch();
    };
```
Should contain the code below so that updateCompletedEvalsCall looks more like this after its handled locally:
```
    const updateCompletedEvals = async () => {
        await updateCompletedEvalsCall({ variables: {
            evalArray: completedEvalsArray
        }});
        refetch();
    };
```

brianpippin: Yes I would think it would clean up the code a lot if you just passed the current array of completed objects, you send it on the get, just send it back on the set modified.  I could be wrong though.

jaudick: I will give it a shot

jaudick: @brianpippin I pushed a commit with the change

brianpippin (approved):

PR #121, opened 2022-11-29
MCS-1554: Add support for imitation goal type

brianpippin:
Thankfully most of the UI is dynamic and not reliant on these goal category types, so few changes.


ThomasSchellenbergNextCentury (approved):
Looks good. Did we already add support for the "passive" test type used by Seeing Leads to Knowing?

PR #128, opened 2023-03-07
MCS-1696 - Split out tool properties for Test Overview roll up

rartiss55:


ThomasSchellenbergNextCentury (commented):
Hey, I can test this now! The UI looks like it's supposed to.

However, I'm wondering if we need to have separate columns for the different directions of MoveObject? Because MoveObject in itself doesn't tell us much. MoveObject forward is similar to PushObject, but MoveObject backward is similar to PullObject. And MoveObject sideways isn't relevant now, but might be in Eval 7.

brianpippin (approved):
The psychs haven't asked us to break out the move action by that much, so I wouldn't want to add more work for ourself until they ask for something.

ThomasSchellenbergNextCentury:
@brianpippin @rartiss55 What specifically did KDL request? Because I doubt they remember the difference between the Move and Push/Pull actions. I agree with Brian's sentiment, but I want to gauge how likely it is that this will become an issue. Seems like it would be easier to address now than having to edit the database post-ingest.

ThomasSchellenbergNextCentury (approved):
We can merge this now if we feel we need it for this week's validation runs, but I'm concerned it will come back to haunt us. (Maybe I'm just being overly concerned.)

jaudick (approved):

PR #131, opened 2023-04-03
MCS-1598 - new slices chart on test overview

rartiss55:
Chart is underneath table on test overview page for slices/keywords view. ~~Also made the keywords clause an "or" instead of an "and" so they can view data across slice levels.~~

### DIFF
--- a/node-graphql/server.statsFunctions.js
+++ b/node-graphql/server.statsFunctions.js
@@ -457,7 +457,11 @@ function processHyperCubeStats(hyperCubeProjection, includeDoNotAnswer, statId, 
     }
 
     statArray.sort((a, b) => (a[statName] > b[statName]) ? 1 : -1);
-
+    // make sure if this is by slice level and slice ends in a number, we sort by that numerical value as well
+    if(sliceType === "level") {
+        statArray.sort((a, b) => /\d+$/.test(a.slice) - /\d+$/.test(b.slice) || a.slice.localeCompare(b.slice, undefined, { numeric: true }));
+    }
+    
     let totalStatObj = getEmptyStatsObject();
     totalStatObj[statName] = "Totals";
 
--- a/node-graphql/server.statsFunctions.js
+++ b/node-graphql/server.statsFunctions.js
@@ -390,6 +390,31 @@ function processStatNameField(statName, sliceLevel) {
     }
 }
 
+function processStatNameKeywords(statName, sliceKeywords) {
+    let sliceKeywordValues = [];
+    for(let i=0; i < sliceKeywords.length; i++) {
+        sliceKeywordValues.push(sliceKeywords[i]["value"]);
+    }
+
+    const containsAll = sliceKeywordValues.every(element => statName.includes(element));
+
+    if(containsAll && sliceKeywordValues.length > 0) {
+        let statStr = "";
+
+        for(let i=0; i < sliceKeywordValues.length; i++) {
+            statStr += sliceKeywordValues[i];
+
+            if(i < sliceKeywordValues.length - 1) {
+                statStr += ", ";
+            }
+        }
+
+        return statStr;
+    } else {
+        return null;
+    }
+}
+
 function getEmptyStatsObject() {
     return {
         "correct_plausible": 0,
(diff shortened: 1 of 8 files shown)


node-graphql/server.statsFunctions.js
```diff
@@ -426,34 +401,47 @@ function getEmptyStatsObject() {
...
-            newStatName = processStatNameKeywords(hyperCubeProjection[i]["_id"][statId], sliceKeywords);
 
-            if(newStatName === null) {
-                continue;
-            }
-            statObj = statArray.find(element => element[statName] === newStatName);
+            statObj = checkForStatObj(statObj, statArray, statName, newStatName, hyperCubeProjection, i);
         }
+    } else {
+        sliceKeywords.forEach(item=> {
```
rartiss55, line 432: In hindsight, wondering if "Totals" doesn't make sense now/needs to be recalculated because we're going across slice levels now?

phile-caci (commented):
@rartiss55 ,

For the keyword "or", I selected two, but the graph doesn't display both of them, it displays if one is selected (let me know if I'm using that wrong).
 
image

ThomasSchellenbergNextCentury:
When I tested this, the graph would be empty if there wasn't any data to show -- maybe that's what's happening to Phi? Not sure. Maybe in the future having a "No Data" text overlay on the chart would be clearer.

You probably already know this, but the current version doesn't have the error lines or the padding between bars that Koleen requested.

rartiss55:
@phile-caci I wasn't seeing that, but its probably moot now because I reverted that change on the keywords view (I think switching from "and" to "or" is messing with the Totals at the bottom and I'm not sure how to reconcile it at the moment).

Since it is working the old way again, if you select keywords from the same slice category, it'll treat it as an "AND", but those slices selected are mutually exclusive, so that's why there's no data.

rartiss55:
> When I tested this, the graph would be empty if there wasn't any data to show -- maybe that's what's happening to Phi? Not sure. Maybe in the future having a "No Data" text overlay on the chart would be clearer.
> 
> You probably already know this, but the current version doesn't have the error lines or the padding between bars that Koleen requested.

I hadn't added the new version yet, should be there now ( it will appear on the Slice Level view only, not keyword). 

Since I nixed the charts from the keywords view, the no data case shouldn't happen, but I did add something in to make sure no chart is displayed if there is no data.

ThomasSchellenbergNextCentury:
@rartiss55 
1. Why does the chart only appear at slice level view? I thought the keyword view was more important for Koleen.
2. The on-hover behavior highlights the left/right padding around the bars. That looked a little weird to me, but not a big deal.

rartiss55:
> @rartiss55
> 
> 1. Why does the chart only appear at slice level view? I thought the keyword view was more important for Koleen.
> 2. The on-hover behavior highlights the left/right padding around the bars. That looked a little weird to me, but not a big deal.

@ThomasSchellenbergNextCentury I had posted a screenshot of graphs with different keywords across slice levels in Slack and got a reply along the lines of "those keywords aren't necessarily in the same batch". I interpreted that to mean Koleen was interested in slices at the same level in terms of the graphs + that's consistent with the graphs she's previously created. She did request the "or" keyword stuff, but I think I will circle back to that as a separate ticket.

ThomasSchellenbergNextCentury:
@rartiss55 There are a bunch of updates from today's KDL meeting. Koleen doesn't expect everything to be done for this eval, but maybe for the next one.

1. She liked how they looked overall.
2. Sounds like having charts on the Level tab is useful, but having them on the Keyword tab would be the MOST useful.
3. She understands the difficulties with getting the filtering working correctly on the Keyword tab. It's a combination of AND and OR. I had her explain it to me with some examples. Long story short, I think we can reduce the configurability a little but keep the workflow she really needs:

Let's say there's a hypercube with 2 slices, each with 3 variables: Color:Red, Color:Green, Color:Blue, Shape:Cube, Shape:Cylinder, Shape:Sphere.

She wants to choose one slice/variable combo, and get the data for all the variables in the other slice. For example, she may choose "Color:Red and Shape:Everything". The filter would look like: Color:Red AND (Shape:Cube OR Shape:Cylinder OR Shape:Sphere). So this would be 3 bars, one for each Shape. Then she may want to choose "Color:Green and Shape:Everything", or "Shape:Cube and Color:Everything", etc.

If the hypercube has 3 or more slices, she'd want to be able to pick a specific variable from 2 slices, and get the data for all the variables from a third. So if we add this slice to our previous example: Position:Left, Position:Right. Then she might want "Color:Red and Shape:Cube and Position:Everything", a.k.a. Color:Red AND Shape:Cube AND (Position:Left OR Position:Right). The chart in this example would have only 2 bars (one for each Position).

(For a hypercube with 4 or more slices, it's OK if the charts can only handle a max of 3 slices at a time, like the example above.)

I think this shares some similarities with what Brian had implemented originally, but Koleen needs to be able to configure the slices, rather than us just using the arbitrary order they're listed in the `slices` array. I'm personally wondering if, instead of the checkboxes, we have 2 or 3 dropdowns, which you use to select the individual slice variables like I've laid out in my examples? (There's plenty to figure out though: each slice would need to have its own "Everything" item in the dropdowns, and you'd have to select exactly one "Everything", and you couldn't select multiple items from the same slice.) Something to chew on.

@brianpippin offered to help with figuring out the queries, and Koleen knows this won't get done in the next 48 hours (especially with you out!). Hopefully this makes a little sense -- Brian should feel free to help fill in the gaps, and Koleen said she'd be happy to talk with you about it more.

ThomasSchellenbergNextCentury:
@rartiss55 Meant to say this during standup: I'm fine with merging this now, and updating the functionality on the keyword tab as a separate ticket. I have no strong opinion either way though.

ThomasSchellenbergNextCentury (approved):

rartiss55:
Thanks for the very helpful comment @ThomasSchellenbergNextCentury -- I copied that over in a separate ticket (MCS-1768).

brianpippin (approved):

PR #133, opened 2023-04-14
MCS-1757: Changes to EvalStatus page for Eval 5 onward

brianpippin:


ThomasSchellenbergNextCentury:
I'm getting console errors when I flip to the eval status page. Is this error reproducible, or does it just seem like a setup issue on my end?

Screenshot from 2023-04-14 12-49-50

brianpippin:
> I'm getting console errors when I flip to the eval status page. Is this error reproducible, or does it just seem like a setup issue on my end?
> 
> Screenshot from 2023-04-14 12-49-50

That console error has been there for awhile and happens one time when loading the page and doesn't seem to effect anything.  I haven't been able to quite locate the source of it yet.  Perhaps a new ticket to investigate.

ThomasSchellenbergNextCentury:
@brianpippin Hmm, well, the page isn't loading for me, and the "groupToProcess is undefined" error on line 250 seems to be something new you added in this PR?

brianpippin:
> @brianpippin Hmm, well, the page isn't loading for me, and the "groupToProcess is undefined" error on line 250 seems to be something new you added in this PR?

I was looking at the top error.  Did you rebuild the back end container as well as the front end container?

ThomasSchellenbergNextCentury:
I wasn't sure if the errors were related.

I rebuilt both the node-graphql and the analysis-ui containers, if that's what you mean.

brianpippin:
> I wasn't sure if the errors were related.
> 
> I rebuilt both the node-graphql and the analysis-ui containers, if that's what you mean.

Indeed, I wonder if it is a condition if there is no status object yet, let me blow away some things locally and try again.

brianpippin:
So if anyone else is testing this.  The first thing you need to do is remove Evaluation 5 Scenes from the "eval_status" collection.  The second thing you need to do is make sure the domain type script in 0.6.5.1 is run as well.  If there is anything we missed we can add it here.

phile-caci (approved):
Looks good on my end:
(the right screen shot)
image

ThomasSchellenbergNextCentury (approved):
Got it working now. Looks good!

Out of curiosity, why is there a separate Configure step? Why doesn't the page come pre-configured?

PR #135, opened 2023-04-18
Fix debug scene link for Eval 6 scenes.

rartiss55:
Saw something about this in Slack, I think this is the right naming pattern for eval 6.

### DIFF
--- a/analysis-ui/src/components/Analysis/sceneDetails.jsx
+++ b/analysis-ui/src/components/Analysis/sceneDetails.jsx
@@ -59,14 +59,6 @@ function SceneDetailsModal({show, onHide, currentSceneNum, currentScene, constan
         setCurrentObjectNum(objectKey);
     }
 
-    const isPostEval4 = (evalSceneName) => {
-        return ["Evaluation 4 Scenes",
-                "Evaluation 3.75 Scenes",
-                "Evaluation 3.5 Scenes",
-                "Evaluation 3 Scenes",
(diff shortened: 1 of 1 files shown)


analysis-ui/src/components/Analysis/sceneDetails.jsx
```diff
@@ -59,27 +59,23 @@ function SceneDetailsModal({show, onHide, currentSceneNum, currentScene, constan
...
+            (currentScene.eval === "Evaluation 6 Scenes" && currentScene.goal.category === 'agents'))) {
             // Goal ID is part of debug file names for Eval 4 except for agency.
             //
             // Goal ID is part of debug file names for Eval 5
-            // for all scenes including agency, assume this won't change
+            // for all scenes including agency
+            //
+            // Goal ID is only part of agency debug file names for Eval 6
             sceneName = sceneName.concat("_", currentScene.goal.sceneInfo.id[0])
         }
```
ThomasSchellenbergNextCentury, line 80: For Rachel's awareness, since she wasn't at standup:

Running the scene generator with the `--eval` flag removes the cell IDs from the scene filenames. (I don't remember why I thought this was a good idea at the time. Maybe to obfuscate information about the hypercube design?)

I forgot to run with `--eval` for the Eval 6 passive agents tasks, as well as for some tasks in past evals (including all of Eval 5, apparently), which is why those filenames contained cell IDs.

brianpippin (approved):

PR #139, opened 2023-05-01
MCS-1786 - update base image in Dockerfiles

rartiss55:
Realized that node-graphql failed to rebuild this morning because the old base image is using a (now) archived version of Debian. Upgraded from node:12 to node:14 (which uses Debian 10 instead of 9).

### DIFF
--- a/node-graphql/Dockerfile
+++ b/node-graphql/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12
+FROM node:14
 
 # Create app directory
 WORKDIR /usr/src/app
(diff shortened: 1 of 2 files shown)


node-graphql/Dockerfile
```diff
@@ -7,7 +7,7 @@ WORKDIR /usr/src/app
 ARG PORT_ARG=9100
 ENV PORT=$PORT_ARG
 
-ARG AMD_MONGO_TOOLS_URL=https://fastdl.mongodb.org/tools/db/mongodb-database-tools-debian92-x86_64-100.3.1.deb
+ARG AMD_MONGO_TOOLS_URL=https://fastdl.mongodb.org/tools/db/mongodb-database-tools-debian10-x86_64-100.7.0.deb
```
rartiss55, line 10: Non-Mac M1 users: I believe this is the correct URL, please let me know if there are issues.

phile-caci (approved):
Looks good.

Removed my old graphql and analysis-ui containers
Rebuilt those containers
Started the containers. website came up without issue

brianpippin (approved):

PR #140, opened 2023-05-02
MCS-1780 - Add weighted toggle for seeing leads to knowing on home page

rartiss55:
Should have weighted/unweighted toggle now for Passive Agency for Eval 6+ on the home page charts, but will only work for seeing leads to knowing. The other passive agency tasks will always be represented as weighted (or paired in this case).

Note that this was branched off of MCS-1786.

### DIFF
--- a/analysis-ui/src/components/Home/chartContainer.jsx
+++ b/analysis-ui/src/components/Home/chartContainer.jsx
@@ -123,6 +123,13 @@ class ChartContainer extends React.Component {
         return this.toUpperFirstLetters(this.props.domainType);
     }
 
+    displayWeightedToggle() {
+        // toggling weighting for passive agency only applies in
+        // Evals 6 and 7
+        return (this.props.domainType.toLowerCase() !== 'passive agents' ||
+            ["Evaluation 6 Results", "Evaluation 7 Results"].includes(this.props.eval.label));
+    }
+
     render() {
         return (
             <div className='chart-home-container'>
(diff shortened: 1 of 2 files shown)


ThomasSchellenbergNextCentury (commented):
I'm wondering if it would be preferable to split SLTK into a separate bar chart. I'm concerned it may be confusing for TA1 to see a toggle that doesn't affect the whole chart. Does anyone else share this concern? Would it be helpful to ask KDL for their opinion?

rartiss55:
> I'm wondering if it would be preferable to split SLTK into a separate bar chart. I'm concerned it may be confusing for TA1 to see a toggle that doesn't affect the whole chart. Does anyone else share this concern? Would it be helpful to ask KDL for their opinion?

My concern with doing things that way is that its splitting a domain type, and I thought it would make it harder to tell if TA1 passed a domain at first glance if it was split into two charts.

ThomasSchellenbergNextCentury:
> My concern with doing things that way is that its splitting a domain type, and I thought it would make it harder to tell if TA1 passed a domain at first glance if it was split into two charts.

Ah, that's a fair point. Seems like one of those things that's going to be a bit weird any way you slice it, so, _shrug_

phile-caci (approved):
Looks good, rebuilt containers (graphql/analysis-ui), passive agents has toggle, and SLTK is showing weighed/unweighted info.

PR #144, opened 2023-05-08
MCS-1429 - Fix for updating csv file links + minor UI fixes

rartiss55:
MCS-1429 specific changes are in generate_csv.py. Messed with different cache-control settings (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) so that the csv file response doesn't get reused. Seems like you need to do it upfront (before you've ever created a file) and on upload for it to work correctly + our current csvs wouldn't have any updates at this point anyway, so I tested by using a separate local test script creating a new test csv file to upload to dev-evaluation-images with the cache-control setting specified, and making minor updates to the csv file each time.

The fixes in the other analysis-ui are minor things like getting rid of console.log / Unique "key" prop errors, and making sure the categories are capitalized in the Scene Analysis view like they are in Test Overview.


ThomasSchellenbergNextCentury (approved):
The code looks good. Seems like one of those things where we won't be completely certain until we try it in production.

PR #146, opened 2023-05-25
MCS-1804 - For results CSV, save performer position/target info from first element in steps array

rartiss55:
Originally tried to do this on the ingest side via collection keys, but I realized that would impact the query builder as well, so made it csv specific.

Now the csv won't have any steps array fields besides the fields in "STEPS_OUTPUT_FIELDS_TO_SAVE", since mongoexport isn't exporting them anyway. To save the first element for all these fields, it just has to be specified explicitly (so "steps.0.output.position.x" instead of "steps.output.position.x").

### DIFF
--- a/node-graphql/csv-scripts/generate_csv.py
+++ b/node-graphql/csv-scripts/generate_csv.py
@@ -8,6 +8,18 @@ import boto3
 EVALUATION_PREFIX = 'csv-db-files/'
 KEYS_INDEX = "collection_keys"
 
+# For the following step array fields, we want to save the
+# value contained in the first step, and ignore other
+# "steps." fields
+STEPS_OUTPUT_FIELDS_TO_SAVE = [
+    "steps.output.position.x",
+    "steps.output.position.y",
+    "steps.output.position.z",
+    "steps.output.rotation",
+    "steps.output.target.position.x",
+    "steps.output.target.position.y",
+    "steps.output.target.position.z"
+]
 
 def upload_csv_file(csv_file_name, bucket_name):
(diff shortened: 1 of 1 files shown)


rartiss55:
Wanted to add: I tested this locally, but in case you'd like to as well, make sure you comment out line 60 in generate_csv.py, since you could upload a file and overwrite what's in S3 (if you have all the most recent data this would be fine, but I definitely do not), and rebuild your node-graphql container. Then you can "docker cp" the csv file and check it. 

Alternatively, we can always just double check things on the dev website once this is merged if the code looks fine before merging to prod, since that's likely easiest :)

phile-caci (approved):
Setup the eval 6 results on my local.
Looks good on my end:

(It took over two hours to generate, is that about normal?)


image


image

rartiss55:
> (It took over two hours to generate, is that about normal?)

Should only take a few minutes to generate. I tried this out with Eval 5 data locally on development as well as this branch, and the times were comparable (only a few minutes). Since it looked like our indexes matched for Eval 6, do you see it taking the same amount of time locally on the development branch?

phile-caci:
> > (It took over two hours to generate, is that about normal?)
> 
> Should only take a few minutes to generate. I tried this out with Eval 5 data locally on development as well as this branch, and the times were comparable (only a few minutes). Since it looked like our indexes matched for Eval 6, do you see it taking the same amount of time locally on the development branch?

yeah, it seems to be taking a little while on development also, probably something on my end then.

ThomasSchellenbergNextCentury (approved):
The code looks good.

PR #148, opened 2023-08-08
MCS-1828set default to sorting for firefox

phile-caci:
Added a default sortby in SortTable.  Seems to default sorting by scene_name ascending in Firefox now.  Retested in Chrome, didn't see any issues.

image


rartiss55 (approved):
Tested sort in Chrome, Firefox, and Safari -- LGTM!

PR #149, opened 2023-10-04
MCS-1956 - Make new distance field visible within query builder

rartiss55:
Minor change needed to see the start_distance_between_performer_and_target field within the UI

Ingest PR here: https://github.com/NextCenturyCorporation/mcs-ingest/pull/139

### DIFF
--- a/node-graphql/server.fieldMappings.js
+++ b/node-graphql/server.fieldMappings.js
@@ -73,7 +73,8 @@ const historyFieldLabelMapTable= {
     "score.weighted_score_worth": "Weighted Evaluation Score Value",
     "score.weighted_confidence": "Weighted Score/Confidence",
     "score.weighted_score": "Weighted Evaluation Score",
-    "scene_goal_id": "Cube/Scene Goal ID"
+    "scene_goal_id": "Cube/Scene Goal ID",
+    "start_distance_between_performer_and_target": "Start Distance Between Performer and Target"
 };
 
 const historyFieldLabelMap = {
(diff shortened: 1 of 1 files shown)


phile-caci (approved):
LGTM:

image

phile-caci:
@rartiss55 , sry, I just noticed that in the Query Builder screen shot it looks like it's showing two of the same record... is that expected?

rartiss55:
> @rartiss55 , sry, I just noticed that in the Query Builder screen shot it looks like it's showing two of the same record... is that expected?

No, that shouldn't happen (unless they had different performers). If you do a query directly against mongo, are you getting the same results?

phile-caci:
> > @rartiss55 , sry, I just noticed that in the Query Builder screen shot it looks like it's showing two of the same record... is that expected?
> 
> No, that shouldn't happen (unless they had different performers). If you do a query directly in mongo, are you getting the same results?

I think there's only one in my mongo

from mongo:
```
mcs> db.eval_6_results.count({"performer": "MESS", "name": "hotel_0001_01", "category_type":"set rotation"})
1
```

image

rartiss55:
> I think there's only one in my mongo
> 
> from mongo:
> 
> ```
> mcs> db.eval_6_results.count({"performer": "MESS", "name": "hotel_0001_01", "category_type":"set rotation"})
> 1
> ```
> 

And is there one record in eval_6_scenes for the scene file as well?

phile-caci:
Thanks @rartiss55 !  Yup, I somehow imported the scenes twice.  Fixed it and now only seeing the one record in the Query builder

phile-caci (approved):
Looks good!

rartiss55:
> It's working for me. Is there a reason why the Query Builder is showing `start_distance_between_performer_and_target` instead of `"Start Distance Between Performer and Target"`?

Because I didn't add a pretty label since I didn't think they'd be querying on it.

PR #150, opened 2023-10-05
MCS-1965 - Adding important tags for Set Rotation in Eval 7

rartiss55:
(Note that this was branched off of MCS-1956).

Once again, if the fields aren't appearing in the scenes CSV file when we actually run the eval, we'll just need to run the update_collection_keys_if_missing.py on the ingest side (I ran this locally to test). Verified that the tags mentioned in this ticket are already saved in ingest properly within the scene record, these minor changes here are just needed so that Koleen will also be able to see + query on these fields in the query builder.

### DIFF
--- a/node-graphql/server.fieldMappings.js
+++ b/node-graphql/server.fieldMappings.js
@@ -484,7 +484,12 @@ const sceneFieldLabelMap = {
     "goal.sceneInfo.targetBehind": "Target Behind",
     "goal.sceneInfo.occludersTrained": "Occluders Trained",
     "goal.sceneInfo.targetHidden": "Target Hidden",
-    "goal.sceneInfo.id": "Cube ID"
+    "goal.sceneInfo.id": "Cube ID",
+    "goal.sceneInfo.containerBaited": "Container Baited",
+    "goal.sceneInfo.containerBaitedIndex": "Container Baited Index",
+    "goal.sceneInfo.landmark": "Landmark",
+    "goal.sceneInfo.landmarkIndex": "Landmark Index",
+    "goal.sceneInfo.landmarkColor": "Landmark Color"
 };
 
 const sceneFieldLabelMapTable = {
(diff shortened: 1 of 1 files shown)


phile-caci (approved):
LGTM!

ThomasSchellenbergNextCentury (commented):
So where are these tags supposed to be? I'm pretty sure I've ingested everything properly (using brand new Set Rotation scenes) but don't see the tags in the Query Builder.

rartiss55:
> So where are these tags supposed to be? I'm pretty sure I've ingested everything properly (using brand new Set Rotation scenes) but don't see the tags in the Query Builder.

They should be at the very end of the table (scroll right) + they would appear as pretty labels in the dropdown for querying fields in "Evaluation x Scenes" (don't know if you went with Evaluation 6 or 7 as your label). 
image

If you aren't seeing them, you either have to rebuild and restart your node-graphql container or run "update_collection_keys_if_missing.py" locally.

ThomasSchellenbergNextCentury:
@rartiss55 I see them in the table, but not in the Field dropdown. I labeled them as Eval 6, and am looking at the Evaluation 6 Scenes. (I'd share a screenshot, but Linux auto-closes my dropdown whenever I take one.)

Here's what I did:
1. Generated a new Set Rotation hypercube with "Evaluation 6 Scenes" as the eval name, and verified the existence of the tags in the debug scene files.
2. Used `local_scene_ingest.py` to ingest the scene files.
3. Ran one of the scenes to generate a history file.
4. Edited the history file to have an `evaluation_name` of `eval_6`
5. Used `local_history_ingest.py` to ingest the history file.

Do I need to restart the docker containers?

ThomasSchellenbergNextCentury (approved):
Got it to work -- I had to run `update_collection_keys_if_missing` **twice**: once on Scenes, once on Results (I think "Scenes" is the relevant one here). Thanks Rachel.

